PR #1, opened 2020-12-29
Got GraphQL Up!

megandouli:
Added prisma schema and fixed a couple prisma import statements.
Apparently an Apollo Server from `apollo-server-express` doesn't have a listen() method, so I switched to using `apollo-server`.


gamburgm (approved):
:+1:

PR #6, opened 2021-01-21
graphql support querying course/section based on coursehash/sectionhash

megandouli:
## why
SearchNEU wants to query for course/sections based on the course hash and section hash.

## what
Added support for graphql to handle these types of queries.

### DIFF
--- a/Keys.js
+++ b/Keys.js
@@ -111,6 +111,21 @@ class Keys {
 
     return hash;
   }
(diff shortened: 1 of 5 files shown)


megandouli:
Error handling necessary for invalid course hash or section hash?

gamburgm (approved):
bet

PR #7, opened 2021-02-04
CoE Scraper Fix

megandouli:
Scrapers for College of Engineering were broken because Jim Johnson from mStoner put himself as a faculty member on the website. See [https://coe.northeastern.edu/people/johnson-jim/](https://coe.northeastern.edu/people/johnson-jim/) and [https://www.mstoner.com/work/case-studies/northeastern-university-college-of-engineering/](https://www.mstoner.com/work/case-studies/northeastern-university-college-of-engineering/) and [https://www.mstoner.com/team/jim-johnson/](https://www.mstoner.com/team/jim-johnson/).

Fix: do some checking on title and img for people before accessing them


scrapers/employees/coe.js
```diff
@@ -47,9 +46,10 @@ class COE {
         obj.link = link;
       }
 
-      let title = $('div.caption').get(0).children[0].data.trim();
-      title = title.replace(/,$/i, '');
+      let title = $('div.caption').get(0) && $('div.caption').get(0).children[0].data.trim();
```
gamburgm, line 49: Might want to try the chaining operator if we support it https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining. Do you think it's worth it to assign a separate variable for the shared expression?

gamburgm (commented):
Looks good, just a small q first, and also, how hard is it to write a test that fails on the old code but succeeds of the new code?

scrapers/employees/coe.js
megandouli: We tried doing `$('div.caption').get(0)?.children[0].data.trim()` but it didn't seem to recognize the operator. What do you mean by a shared expression?

megandouli: actually it looks like optional chaining works but ESLint doesn't like it

PR #9, opened 2021-02-05
Ad dump processor fix

andrewydai:


dumpProcessor.ts
```diff
@@ -130,12 +131,12 @@ class DumpProcessor {
...
-      return prisma.course.update({
+    await pMap(Object.entries(courseUpdateTimes), async ([id, updateTime])=> {
+      await prisma.course.update({
         where: { id },
         data: { lastUpdateTime: updateTime },
       });
-    }));
+    }, {concurrency: 10})
 
     macros.log('finished updating times');
```
megandouli, line 142: Can you also change the one below (inserting subjects) to use pmap instead of Promise.all?

PR #12, opened 2021-03-09
Add Sections For ClassOccurrence Resolvers

megandouli:
# what 
For class pages, we want to make the `class(subject: String!, classId: String!)` graphQL query instead of the standard `search` query. The resolvers on the ClassOccurrences were not returning Section info, causing the desired query to get an error response.

# how 
Add `include: { sections: true }` in prisma in all the resolvers that return a ClassOccurrence.

# additional issues/questions
The GraphQL typedef for Sections includes a `termId` field but the Sections in ClassOccurrences returned from resolvers do not have this field.

### DIFF
--- a/graphql/resolvers/class.ts
+++ b/graphql/resolvers/class.ts
@@ -2,35 +2,45 @@
  * This file is part of Search NEU and licensed under AGPL3.
  * See the license file in the root folder for details.
  */
-import prisma from '../../services/prisma';
-import HydrateCourseSerializer from '../../serializers/hydrateCourseSerializer';
-import keys from '../../utils/keys';
+import prisma from "../../services/prisma";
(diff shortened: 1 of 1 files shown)


graphql/resolvers/class.ts
```diff
@@ -2,62 +2,87 @@
...
   return serializeValues(results);
-}
+};
 
 const getClassOccurrence = async (termId, subject, classId) => {
   const res = await prisma.course.findUnique({
     where: {
       uniqueCourseProps: { subject, classId, termId },
     },
+    include: { sections: true },
```
andrewydai, line 39: I'm not too familiar with backend setup, maybe Dillon's linter story fixes this but i think having some standard formatting like a prettierrc file could help make these file changes a lot smaller so it's easier to know what's actually changed (from my understanding it's just the `include: { sections: true }`. Otherwise looks good :)

PR #14, opened 2021-03-14
Scrape course fees

dillonhammer:
# what 
Course fees were not being scraped, we want to scrape course fees.

# how
Hit the `/getFees` banner endpoint to scrape course fees and insert into database.


megandouli (approved):
Is there a reason for the empty file `create-multiple-postgresql-databases.sh`? otherwise looks great!

dillonhammer:
@megandouli I had to change the permissions on that file in order to run `yarn dev:docker`, not sure how it was working for people before that.

megandouli:
@dillonhammer I'm leaning towards leaving that out of this PR for now, testing if this is an issue on other people's machines, and making a separate PR to fix it if it is

PR #17, opened 2021-03-19
Custom scrapes

dillonhammer:
# What
We want to speed up the scraping process so that scraper-related dev work doesn't get bottle-necked by 10+ minute scrapes.

# How
Implement custom scrapes, where we can specify some rules to scrape only a subset of all courses. Documentation has been added to the README.


scrapers/classes/parsersxe/bannerv9Parser.js
```diff
@@ -24,6 +28,16 @@ class Bannerv9Parser {
...
     macros.log(termsUrl);
+
+    // If scrapers are simplified then this logic would ideally be moved closer to the scraper "entry-point"
+    if (process.env.CUSTOM_SCRAPE && filters.truncate) {
+      macros.log("Truncating courses and sections tables");
+      const clearCourses = prisma.course.deleteMany({});
+      const clearSections = prisma.section.deleteMany({});
+      await prisma.$transaction([clearCourses, clearSections]);
+      macros.log("Truncating elasticsearch classes index");
+      await elastic.resetIndex(elastic.CLASS_INDEX, classMap);
```
megandouli, line 39: why do we have to reset the elasticSearch indices?

megandouli (approved):
ahmazing the scrapes work just as described in the docs! just a small question about elasticSearch since I noticed `elastic.resetIndex` doesn't seem to be called anywhere except via a `package.json` command

scrapers/classes/parsersxe/bannerv9Parser.js
dillonhammer: This is so that if you set `truncate` to true then it will clear out both Postgres and elasticsearch before repopulating with the custom scrape. I was thinking that if we're going to truncate one database it would make sense to truncate them all.

PR #19, opened 2021-03-20
Fix Scrapers For Min/Max Credits

megandouli:
# what 
Scrapers were reversing minCredits and maxCredits

# fix
If minCredits and maxCredits are the same, scrapers will get 0 for `creditHourHigh` and the number of credits for `creditHourLow`. Therefore `maxCredits` should be `creditHourHigh` if it exists and `creditHourLow` otherwise.


NEUDitao (approved):
🙄

PR #22, opened 2021-04-03
Fix How Updater Sends NotificationInfo to Notifyer

megandouli:
# what 
The notification info wasn't being sent in the request to Search's notifyer.

# how 
Just do `req.end(body)` instead of `req.write(body)`. Also for local dev work, the request should be http not https.

# IMPORTANT!! notes on merging this into prod 
- If there isn't one already, we need a `WEBHOOK_PRIVATE_KEY` environment variable and `UPDATER_URL` on prod

## how to generate public and private key
Run this command in your terminal `ssh-keygen -m PEM -t rsa -b 4096 -C "youremail@example.com"`, type in the name of the file you want the keys to be in, and your private key will be named `yourFileName.pub` and your public key will be named `yourFileName`

### DIFF
--- a/services/updater.ts
+++ b/services/updater.ts
@@ -7,6 +7,7 @@ import _ from "lodash";
 import pMap from "p-map";
 import { Course, Section } from "@prisma/client";
 import * as https from "https";
+import * as http from "http";
 import * as httpSignature from "http-signature";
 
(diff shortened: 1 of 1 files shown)


gamburgm (approved):
👍 yeet (let's not merge until we've moved secrets over)

PR #23, opened 2021-04-08
Deploy cleanup

dillonhammer:
# what

The most recent prod deployment exposed a few bugs, these fixes should get master into a state where it is able to build a working docker image.

# how

- Run a `yarn install --production` in the dist folder instead of just copying over node_modules.
- Try/catch the prisma client generation, since that was failing silently prior to the quick fix.
- Copy over the `prisma/.env` file, since the Dockerfile copy was ignoring that earlier and causing prisma client to fail.
- Add license to package.json to get rid of warning printout when running CCA
- Clear out the dist folder before building, and delete the .git folder from dist after building


gamburgm (approved):
big chungus

PR #25, opened 2021-04-11
Change Instance Type of ES on Staging

gamburgm:
read title


megandouli (approved):
cool just wanna remove the load balancer changes

PR #29, opened 2021-04-13
Copy scripts folder in Dockerfile

dillonhammer:
# what
The updater was exiting early on AWS, we want to fix that.

# how
The Dockerfile wasn't copying the `scripts` folder, so we add a line to copy it.


megandouli (approved):
awesome

PR #32, opened 2021-04-14
Removing Updater Logs That Were Added For Debugging Prod

megandouli:
Previously added some logs to see what private webhook key, updater URL were. No longer need this so this PR is removing some of those logs.


gamburgm (approved):
Bet

PR #39, opened 2021-10-02
Add resetting ES index in scrapers/main

hankewyczz:
**Problem:** Right now, there's only a manual script for resetting the ElasticSearch index that we have to remember to run on new environments. This has resulted in prod sometimes not recognizing queries for "fundies" and "software dev".

**Solution:** Reset the ElasticSearch index when we run the scrapers.

This assumes that we always want to reset the indexes whenever the main scrapers are run.


kiminkim724 (approved):
Nice!

scrapers/main.js
```diff
@@ -8,15 +8,28 @@ import macros from "../utils/macros";
...
 
 // Main file for scraping
 // Run this to run all the scrapers
 
 class Main {
   async main() {
     const classesPromise = classes.main(["neu"]);
-
-    const promises = [classesPromise, matchEmployees.main()];
+    const elClassPromise = elastic.resetIndex(elastic.CLASS_INDEX, classMap);
```
megandouli, line 21: personally would prefer to name this `esResetClassPromise` and `esResetEmployeePromise` which I think matches some other code in CCA where we `generateESQuery`

megandouli (changes requested):
don't merge yet

PR #41, opened 2021-10-03
Delete Non-Existent Sections / Courses on scrapes 

hankewyczz:
https://trello.com/c/fs503gwU/241-process-for-deleting-non-existent-sections-courses

### DIFF
--- a/scrapers/main.js
+++ b/scrapers/main.js
@@ -20,7 +20,11 @@ class Main {
 
     const [termDump, mergedEmployees] = await Promise.all(promises);
 
-    await dumpProcessor.main({ termDump: termDump, profDump: mergedEmployees });
+    await dumpProcessor.main({
+      termDump: termDump,
+      profDump: mergedEmployees,
+      destroy: true,
+    });
 
     macros.log("done scrapers/main.js");
   }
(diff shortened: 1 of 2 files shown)


megandouli (approved):
nice! way to catch this option for the dump processor!

hankewyczz:
Added support for deleting old sections:

- Schema update to include a `last_update_time` for sections
- Schema update for both Sections and Courses to make `last_update_time` not optional, and to provide a default value
- Serializer update for Sections to properly parse the Datetime (I don't think this is needed anymore, but Courses does the same thing so I kept it for consistency)
- Remove sections if they were last seen over 2 days ago, and if the TermID of the associated Course is one of the active ones.

megandouli (approved):
great work!

graphql/tests/search.test.seq.ts
```diff
@@ -29,6 +29,8 @@ const FUNDIES: Course = {
   minCredits: 0,
   coreqs: EMPTY_REQ,
   prereqs: EMPTY_REQ,
+  feeAmount: 0,
+  feeDescription: "",
```
andrewydai, line 33: Was this a change you made?

PR #45, opened 2021-10-24
add hardcoded term id's to updater

andrewydai:
This adds hardcoded term id's to the updater for spring undergrad,  and spring, fall, and winter CPS/LAW semesters/quarters.


megandouli (changes requested):
you can remove the `2021`'s we don't need to keep updating summer courses

PR #48, opened 2021-11-01
Fix HTML entity encoding rendering

hankewyczz:


megandouli (changes requested):
any way you can add some tests for this?

megandouli (approved):
awesome, I wanna test this out locally before merging but lgtm!

andrewydai (approved):

PR #49, opened 2021-11-02
add reindexing script and refactor of elastic.ts

andrewydai:
This is a second release of this [pr](https://github.com/sandboxnu/course-catalog-api/pull/39). Essentially, we realized that dropping the whole ES index is not ideal. 

The solution is to refactor our ES indexes to follow a blue-green deployment. This is a little more involved than I thought and the amount of code change sort of reflects that. The changes essentially boil down to this:

- Index names are now dynamically set as either blue or green. Everything that interacts with our indexes can still refer to them by their base name, i.e. "classes" instead of "classes_blue". However, under the hood their names now have blue or green attached to them. Thus, all references to index names are now referencing index aliases.
- Any time an index is created, it must now be aliased as a base name like "classes"
- The original resetIndex script is now a ts file, and does not specify what to reset. Instead, all the indexes are reset.
- I've removed resetting the index anywhere in the bannerv9parser, since I think it's too risky to use it in the parser. This script can only be invoked by the CL now.
- I've added a new script resetIndex:noLoss (yeah the naming sucks not sure what to put). This script essentially creates a second index, moves the data over, then switches the alias and drops the first index. Read this [blog](https://hamidmosalla.com/2021/01/30/how-to-change-elastic-search-index-mapping-without-losing-data/) for more info
- CRITICAL 1: There is a critical limitation for this. Basically, a big reason you want to reset your index with the new script is because you have a mapping change, but you don't want to lose your data. However, reindexing (moving data from one index to another) is not that smart. If the two indexes differ in mapping, it may be problematic. The list of things that will work smoothly in terms of differences are: Data type changes if compatible, adding a field, removing a field. Anything more complex requires a script to map the data while reindexing, which would have to be inline in our case. I tried to think of a solution that was clean but couldn't and wanted to push this out first.
- CRITICAL 2: ES is not transactional, meaning we don't have guaranteed validity of the transferred data. Right now I have no protocol for checking that the reindex validly transferred data other than no error occurring. In general, this is kind of the point of ES (trade safety for speed) but unsure if we want to be more rigorous on a check.


scripts/resetIndex.ts
```diff
@@ -1,15 +1,12 @@
...
   macros.log("Resetting indices for URL ", macros.getEnvVariable("elasticURL"));
   (async () => {
-    await elastic.resetIndex("classes", classMap);
-    await elastic.resetIndex("employees", employeeMap);
+    await elastic.resetIndex();
     macros.log("Success! Closing elastic client and exiting.");
     elastic.closeClient();
     process.exit();
-  })().catch((e) => macros.error(e));
+  })().catch((e) => macros.log(e));
```
megandouli, line 11: any reason you chose `macros.log` instead of `macros.error`?

utils/elastic.ts
```diff
@@ -15,22 +15,58 @@ import {
...
+  private initializing: Promise<void>;
 
   constructor() {
     // Because we export an instance of this class, put the constants on the instance.
-    this.CLASS_INDEX = "classes";
-    this.EMPLOYEE_INDEX = "employees";
+    this.CLASS_ALIAS = "classes";
+    this.EMPLOYEE_ALIAS = "employees";
+
+    this.initializing = this.initializeIndexes();
```
megandouli, line 47: so to make sure I understand this correctly, in the constructor, we're creating a promise to initialize indices then in the reset methods, we need to wait for this promise to resolve? I don't see the purpose of line 47 when we could just do `await this.initalizeIndexes()` in the reset methods instead. It also seems brittle that now we need to remember to await this promise at the start of every method

megandouli (commented):
some questions... just wanna go through them before approving!

utils/elastic.ts
andrewydai: Yeah we could do it only in the reset methods too, I hadn't thought about that. I just didn't want to make a request to get the current index names each time you want to run those methods, but realistically shouldn't end up mattering since you only really run those scripts once.

scripts/resetIndex.ts
andrewydai: Uh no I didn't mean to change that tbh don't know how it got changed

megandouli (approved):
yay

utils/elastic.ts
```diff
@@ -43,33 +79,148 @@ export class Elastic {
...
+          source: {
+            index: name,
+          },
+          dest: {
+            index: nextIndexName,
+          },
+        },
+        wait_for_completion: false,
+      });
+
```
megandouli, line 161: just add a comment explaining this is polling ES and why it's a necessary step

PR #50, opened 2021-11-07
Remove hardcoded term IDs

hankewyczz:
https://trello.com/c/xRH6xepp/216-remove-hardcoded-termids


prisma/migrations/20211024221117_reset_term_ids/migration.sql
```diff
@@ -0,0 +1,10 @@
+/*
+  Warnings:
+
+  - Added the required column `sub_college` to the `term_ids` table without a default value. This is not possible if the table is not empty.
+  - Added the required column `text` to the `term_ids` table without a default value. This is not possible if the table is not empty.
+
+*/
+-- AlterTable
+ALTER TABLE "term_ids" ADD COLUMN     "sub_college" TEXT NOT NULL,
```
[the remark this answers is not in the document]
hankewyczz: Merged

scrapers/classes/parsersxe/bannerv9Parser.js
```diff
@@ -47,8 +47,87 @@ class Bannerv9Parser {
...
+      filterdTermInfos.push(fullTermInfoList[termId]);
+    }
+
+    const allIds = filterdTermInfos.map((termInfo) => termInfo["termId"]);
+    // Delete the old terms (ie. any terms that aren't in the list we pass this function)
+    await prisma.termInfo.deleteMany({
+      where: {
+        termId: { notIn: Array.from(allIds) },
+      },
+    });
```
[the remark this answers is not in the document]
hankewyczz: 1. Thanks for catching that - it use to be an object, not a list. I updated it
2. Any terms in the TermInfo table which no longer have data

ie. If we once had data for a term, but have since lost/removed that data, there's no point in keeping that term (since we can't serve any data for it)

I added a LOT of comments to the function, lmk if it makes sense now.

services/updater.ts
```diff
@@ -28,32 +31,33 @@ interface OldData {
...
-
   SECTION_MODEL: ModelName;
-
   SEMS_TO_UPDATE: string[];
 
-  static create() {
-    return new this();
+  // produce a new Updater instance
+  static async create() {
+    // Scrapes a list of terms IDs (and associated info) from Banner - these are the only ones we want to update
```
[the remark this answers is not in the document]
hankewyczz: Good point - the updater will never insert new courses, only update existing ones, so there's no need to query Banner for it. 

I have to say, I'm not a huge fan of the scraper logic (specifically, when a new term comes out). Worst case, we could wait a week before the scrapers pick up the new term. Just an idea, but:

- The scraper could run every hour, and just query the Banner terms endpoint
- If the newest term from Banner is in our database, do nothing. Wait for the weekly scrape
- If the newest term from Banner is not in our database, that means a new term has been released and we don't have it. Run a full scrape, then return to the weekly schedule. 

Thoughts?

scrapers/classes/parsersxe/bannerv9Parser.js
```diff
@@ -47,8 +70,85 @@ class Bannerv9Parser {
...
+
+    const termInfoList = termList
+      // Sort by descending order (to get the most recent term IDs first)
+      .sort((a, b) => b.termId - a.termId);
+
+    return termInfoList;
+  }
+
+  /**
+   * Given a list of TermIDs, it updates the term_info table
```
megandouli, line 89: would be nice if purpose statement is explicit about how it updates the term_info table

scrapers/classes/parsersxe/bannerv9Parser.js
```diff
@@ -47,8 +70,85 @@ class Bannerv9Parser {
...
+    - A termID is a string (eg. '202230')
+    - A TermInfo is an object, containing the keys:
+      - 'termId' (which is a termID)
+      - 'subCollegeName' - the name of the college associated with this term
+      - 'text' - an English, textual description of this term (eg. 'Spring 2021 Semester')
+    */
+
+    // Get a list of termIDs for which we already have data
+    //  (ie. terms we've already scraped AND that still have courses associated with them)
+    let existingIds = await prisma.course.groupBy({ by: ["termId"] });
```
megandouli, line 104: just make it a `const` one-liner instead of a `let` then reassigning it on the next line:

```
const existingIds = (await prisma.course.groupBy({ by: ["termId"] })).map(t => t.termId);
```

scrapers/classes/parsersxe/bannerv9Parser.js
```diff
@@ -47,8 +70,85 @@ class Bannerv9Parser {
...
+      NUMBER_OF_TERMS_TO_UPDATE
+    );
+
+    // Convert each termID in the list of existingIds to a TermInfo
+    for (const termId of existingIds) {
+      // We query the list of all TermInfo objects to get the one associated with this termID
+      const termInfo = fullTermInfoList.filter((termInfo) => {
+        return termInfo["termId"] === termId;
+      });
+      // Make sure we have a TermInfo for this term
```
megandouli, line 120: it seems like we should really be using `.find` instead of `.filter` then checking the length of the filtered array is > 0.

scrapers/classes/parsersxe/bannerv9Parser.js
```diff
@@ -47,8 +70,85 @@ class Bannerv9Parser {
...
+      if (termInfo.length > 0) {
+        filteredTermInfos.push(termInfo[0]);
+      }
+    }
```
megandouli, line 124: this is a pretty big method... I think you can pull lines 93-124 into a separate method that returns `filteredTermInfos` then use that below to update the database.

scrapers/classes/parsersxe/bannerv9Parser.js
```diff
@@ -47,8 +70,85 @@ class Bannerv9Parser {
...
+    */
+
+    // Get a list of termIDs for which we already have data
+    //  (ie. terms we've already scraped AND that still have courses associated with them)
+    let existingIds = await prisma.course.groupBy({ by: ["termId"] });
+    existingIds = existingIds.map((t) => t["termId"]);
+
+    // Filter the full list of TermInfos to get the terms that we are currently scraping/updating
+    //  This is a subset of all of the terms (usually, we're only scraping ~10 at a time)
+    const filteredTermInfos = fullTermInfoList.slice(
```
megandouli, line 109: honestly still pretty confused by what `filteredTermInfos` represents, is there not a better name for it? Is it just the terms we're currently updating plus terms we have courses for in our database?

services/updater.ts
```diff
@@ -28,32 +30,29 @@ interface OldData {
...
-      "202230",
-      "202214",
-      "202215",
-      "202225",
-      "202234",
-      "202235",
-      "202212",
-      "202232",
-    ];
+    this.SEMS_TO_UPDATE = termIds;
```
megandouli, line 55: you'll also need to update `updater.test.seq.ts` since we hard coded `SEMS_TO_UPDATE` there

megandouli (changes requested):
this is awesome work

scrapers/classes/parsersxe/termListParser.js
```diff
@@ -32,11 +39,11 @@ class TermListParser {
   determineSubCollegeName(termDesc) {
     if (termDesc.includes("CPS")) {
       return "CPS";
-    }
-    if (termDesc.includes("Law")) {
+    } else if (termDesc.includes("Law")) {
       return "LAW";
+    } else {
+      return "None";
```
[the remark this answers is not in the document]
hankewyczz: https://github.com/sandboxnu/course-catalog-api/blob/master/scrapers/classes/parsersxe/termListParser.js#L11-L14

The original term parser left the subcollege as undefined for undergrad, but we needed a string for insertion into the database, so we chose "None" (to keep it close to the prior 'undefined' behavior). Switching to "NEU" would be clearer - we just weren't sure if we wanted to keep the 'undefined' behavior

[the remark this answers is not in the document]
megandouli: I'm good with switching to "NEU"

prisma/schema.prisma
```diff
@@ -124,3 +124,10 @@ model User {
   FollowedCourse FollowedCourse[]
   @@map(name: "users")
 }
+
+model TermInfo {
+  termId          String   @unique @map(name: "term_id")
+  subCollege      String   @map(name: "sub_college")
```
megandouli, line 130: hmmm now that I think about it more, is there any reason we want to store "subCollege" and "text" in our database? it should be possible to derive those values from just the termId, right? Last digit tells you the subCollege and whether it's quarter or semester and second to last digit tells you the season

scrapers/classes/parsersxe/termListParser.js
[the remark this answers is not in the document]
megandouli: also just realized we can determine the subcollege in a less hacky way - just use the last digit of the termID

scrapers/main.js
```diff
@@ -26,6 +25,13 @@ class Main {
       destroy: true,
     });
 
+    macros.log("Updating term IDs");
+
+    // Update the stored term IDs
+    const allTermInfos = bannerv9parser.getAllTermInfos(bannerv9CollegeUrls[0]);
```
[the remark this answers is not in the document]
megandouli: move to dump processor main func

services/dumpProcessor.ts
```diff
@@ -429,6 +430,35 @@ class DumpProcessor {
...
+
+  // Updates the termInfo table - adds/updates current terms, and deletes old terms for which we don't have data
+  async updateTermInfos(termInfos: TermInfo[]): Promise<void> {
+    const termIds = termInfos.map((t) => t.termId);
+
+    // This deletes any termID which doesn't have associated course data
+    //    For example - if we once had data for a term, but have since deleted it, this would remove that termID from the DB
+    await prisma.termInfo.deleteMany({
+      where: {
+        termId: { notIn: Array.from(termIds) },
```
megandouli, line 442: why do we have array.from here?

services/dumpProcessor.ts
```diff
@@ -429,6 +430,35 @@ class DumpProcessor {
...
+      },
+    });
+
+    // Upsert new term IDs, along with their names and sub college
+    for (const { termId, subCollege, text } of termInfos) {
+      await prisma.termInfo.upsert({
+        where: { termId: termId },
```
megandouli, line 449: kv pairs can be simplified to the common name

services/updater.ts
```diff
@@ -28,32 +30,28 @@ interface OldData {
...
+    });
+
+    const termIds: string[] = termInfos.map((t) => t.termId);
+
+    return new this(termIds);
   }
 
-  // DO NOT call the constructor, instead use .create
-  constructor() {
+  // The constructor should never be directly called - use .create()
```
megandouli, line 50: add comment explaining why this constructor can't be private (but still shouldn't be used)

graphql/resolvers/class.ts
```diff
@@ -5,34 +5,38 @@
...
     include: { sections: true },
     orderBy: { termId: "desc" },
   });
   return serializeValues(results)[0];
 };
 
-const getAllClassOccurrences = async (subject, classId) => {
-  const results = await prisma.course.findMany({
+const getAllClassOccurrences = async (subject: string, classId: string) => {
+  const results: Course[] = await prisma.course.findMany({
```
megandouli, line 26: these type annotations aren't accurate... ideally we'd be able to use prisma generated types but it seems to have stopped working and I believe the fix is upgrading our version of typescript. This `Course` type you imported does not have a sections field but you can see on line 28 the returned object would have sections. For now, just remove these types.

megandouli (approved):
amazing work, just one small change then it's good to merge!

twilio/server.ts
```diff
@@ -36,7 +36,7 @@ app.post("/sms/signup", (req, res) => {
       res.status(response.statusCode).send(response.message);
       return;
     })
-    .catch((e) =>
+    .catch(() =>
```
andrewydai, line 39: why is this part of this pr lol

andrewydai (approved):

twilio/server.ts
hankewyczz: it was just a linting thing (unused variable)

PR #51, opened 2021-11-18
Documentation for deployment and services 

megandouli:
# what
Additional docs for deployment


documentation/services.md
```diff
@@ -0,0 +1,22 @@
...
+CCA uses the folling 3rd party services. The site depends on most of these being up to run correctly. Some are just used for analytics.
+
+## Notifications
+
+- Twilio
+
+## Deployment
+
+- Terraform and AWS for deployment
+- Cloudflare for Domain Name System (DNS) management
```
bugsalexander, line 12: i think dns is managed by terraform also–just learned this today from @gamburgm , actually! https://github.com/sandboxnu/dns

megandouli: I'm pretty sure it's Cloudflare for SearchNEU... we're on Ryan Hughes's account still

bugsalexander: oh yes you are so right my bad sorry

documentation/services.md
```diff
@@ -0,0 +1,22 @@
+# Services
+
+CCA uses the folling 3rd party services. The site depends on most of these being up to run correctly. Some are just used for analytics.
```
bugsalexander, line 3:
```suggestion
CCA uses the following 3rd party services. The site depends on most of these being up to run correctly. Some are just used for analytics.
```

PR #52, opened 2021-11-19
Enable Delete Cascade

megandouli:
# What 

Deleting stale courses / sections was breaking because prisma doesn't do cascade delete and in prod, there were records in `followed_courses` and `followed_sections` that point to stale courses / sections.

# How

prisma 3.x.x supports delete cascade in the schema but this would require a migration from prisma 2.x.x and there are some breaking changes that I don't want to figure out now since it's getting annoying to keep deleting the ECS updater task. This PR enables a preview feature for referential actions and updates the schema to handle delete cascade.

## Miscellaneous

TODOs: upgrade prisma to 3.x.x and also typescript so prisma client types work again.


hankewyczz (commented):
Looks good!

PR #53, opened 2021-11-21
Update HE dependency, fix tests

hankewyczz:
Changes:
- Switched from `html-entities` to `he`, since we already use the latter in other serializers
- Updated tests to make them work with `he`
- Updated classParser tests to actually run (most of them were disabled - I'm guessing someone disabled them for testing and never re-enabled them)


puneetjohal (approved):
What a monster commit :0

megandouli (changes requested):
delete package-lock.json and make sure you're installing dependencies using `yarn` only and not `npm`

megandouli (approved):
awesome!

PR #55, opened 2021-11-22
CPS/LAW term filter

SimChadha:
Added term halves for each course to know if they run in the first half, the second half, or the full term. A term half attribute is added to each section in the database after being scraped.


scrapers/classes/parsersxe/sectionParser.js
```diff
@@ -32,6 +33,75 @@ class SectionParser {
...
+          35: ["04", "05"],
+          54: ["05", "06"],
+          52: ["05", "06"],
+          55: ["07", "08"],
+        };
+
+        if (
+          firstHalfDatesMap[suffix][0] ===
+          SR.meetingsFaculty[0].meetingTime.startDate.substring(0, 2)
+        ) {
```
megandouli, line 86: it'd be helpful to declare constants for things like `firstHalfDatesMap[suffix][0]` and `SR.meetingsFaculty[0].meetingTime.startDate.substring(0, 2)` and give those a helpful name. Maybe even a helper function that takes `SR` and returns the month of the start date and end date or something

scrapers/main.js
```diff
@@ -19,8 +19,16 @@ class Main {
...
-    await dumpProcessor.main({ termDump: termDump, profDump: mergedEmployees });
+    console.log("Finished Scraping");
+    try {
+      await dumpProcessor.main({
+        termDump: termDump,
+        profDump: mergedEmployees,
+      });
+    } catch (error) {
+      console.log("dump processor failed");
+      console.log(error);
```
megandouli, line 30: use `macros.log` and `macros.error`

scrapers/main.js
```diff
@@ -19,8 +19,16 @@ class Main {
     const promises = [classesPromise, matchEmployees.main()];
 
     const [termDump, mergedEmployees] = await Promise.all(promises);
-
-    await dumpProcessor.main({ termDump: termDump, profDump: mergedEmployees });
+    console.log("Finished Scraping");
```
megandouli, line 22: use `macros.log`

megandouli (changes requested):

megandouli (commented):
also merge from master and resolve conflicts on your branch first before merging back into master

scrapers/classes/parsersxe/sectionParser.js
```diff
@@ -32,6 +33,75 @@ class SectionParser {
...
+
+      - CPS Summer Semester (54): May - Jun
+      - Law Summer Semester (52): May - Jun
+      - CPS Summer Quarter (55): July - Aug
+
+      Note: law quarters don't have half semesters
+      */
+
+        // maps a termid suffix to a list containing the start and end month for the first half of that term
+        const firstHalfDatesMap = {
```
hankewyczz, line 70: Same as Megan's comment above - all of the term variables should be maintained in a single file, just to have a single source of truth. I think it'd be worth creating a top-level `vars` directory, and create a file underneath it for the terms (there are other things that should probably be global variables, so it allows us to expand in the future).

scrapers/classes/parsersxe/sectionParser.js
hankewyczz: Adding on to this - a good place for this may be `scrapers/classes/parsersxe/meetingParser`. That's where we do most of the meeting/time parsing, so adding a function there to get the month from an SR object might not be a bad idea.

scrapers/classes/parsersxe/sectionParser.js
```diff
@@ -58,6 +128,7 @@ class SectionParser {
         `?term_in=${SR.term}&crn_in=${SR.courseReferenceNumber}`,
       profs: SR.faculty.map(MeetingParser.profName),
       meetings: MeetingParser.parseMeetings(SR.meetingsFaculty),
+      termHalf: this.getTermHalf(SR, SR.term),
```
hankewyczz, line 131: Should `getTermHalf` maybe be in `MeetingParser`? My logic is:
- This file (where it currently is) doesn't have a lot of parsing methods (imports most of them), so I'd prefer to keep that consistent
- Some of the other parsers (like for profs, meetings, etc) already have parsers in `MeetingParser`, and I feel like getting the term half is related enough to meetings where it justifies the connection

hankewyczz (changes requested):

PR #56, opened 2021-11-30
Improve Updater

megandouli:
# what
The updater process currently hangs in prod for some unknown reason. This is annoying because then I have to remember to kill the ECS task daily to spawn a new one that works for a little while. This PR adds some logging and kills the process if the updater errors instead of attempting to run the updater again.

### DIFF
--- a/services/updater.ts
+++ b/services/updater.ts
@@ -66,6 +66,7 @@ class Updater {
         this.update();
       } catch (e) {
         macros.warn("Updater failed with: ", e);
+        process.exit(1); // if updater fails, exit the process so we can spin up a new task and not hang
       }
     }, intervalTime);
     this.update();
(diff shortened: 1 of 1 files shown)


services/updater.ts
```diff
@@ -106,7 +113,9 @@ class Updater {
     );
 
     macros.log(
-      `Done running updater onInterval. It took ${totalTime} ms. Updated ${sections.length} sections.`
+      `Done running updater onInterval. It took ${totalTime} ms (${
+        totalTime / 60000
+      } minutes). Updated ${sections.length} sections.`
```
kiminkim724, line 118: Just wondering why we print the totalTime in both ms and minutes? Is there a specific reason?

megandouli: I'm just curious about the time in minutes and don't want to do the mental calculation myself lmao not sure what the behavior of the updater would be if the process took longer than 5 minutes

kiminkim724 (approved):
Looks good!

PR #57, opened 2021-11-30
properly await update, catch error, exit on error

megandouli:
# what 
Updater hangs in prod and errors weren't being caught.

# how 
Properly await the `update()` method, catch error, and exit the process on error. This will force AWS to spin up a new ECS task for the updater.

### DIFF
--- a/services/updater.ts
+++ b/services/updater.ts
@@ -61,15 +61,20 @@ class Updater {
     // In dev the cache will be used so we are not actually hitting NEU's servers anyway.
     const intervalTime = macros.PROD ? 300000 : 30000;
 
-    const interval = setInterval(async () => {
-      try {
-        await this.update();
-      } catch (e) {
-        macros.warn("Updater failed with: ", e);
-        clearInterval(interval);
-        process.exit(1); // if updater fails, exit the process so we can spin up a new task and not hang
-      }
+    setInterval(() => {
+      this.updateOrExit();
     }, intervalTime);
+
+    this.updateOrExit();
+  }
+
+  async updateOrExit(): Promise<void> {
+    try {
+      await this.update();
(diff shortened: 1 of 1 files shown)


services/updater.ts
```diff
@@ -61,15 +61,15 @@ class Updater {
...
       try {
-        this.update();
+        await this.update();
       } catch (e) {
         macros.warn("Updater failed with: ", e);
+        clearInterval(interval);
         process.exit(1); // if updater fails, exit the process so we can spin up a new task and not hang
       }
     }, intervalTime);
-    this.update();
```
hankewyczz, line 72: Doesn't the first setInterval execution comes after the delay? (so now, the updater won't run when you initialize it, only after the delay).

services/updater.ts
```diff
@@ -61,15 +61,15 @@ class Updater {
...
     const intervalTime = macros.PROD ? 300000 : 30000;
 
-    setInterval(() => {
+    const interval = setInterval(async () => {
       try {
-        this.update();
+        await this.update();
       } catch (e) {
         macros.warn("Updater failed with: ", e);
+        clearInterval(interval);
```
hankewyczz, line 69: So I don't really understand why/if this is needed - shouldn't process.exit() take precedence over the setInterval? (if it works, it works - I'm just curious)

megandouli: yup you're right, I just have a habit of clearing timeouts and intervals bc of software dev rn lol

services/updater.ts
megandouli: correct, I didn't want to call this.update() without wrapping it in a try catch, but I've pushed a new change to solve that now

hankewyczz (approved):
Looks good!

PR #59, opened 2021-12-02
Fix Updater Error If User Unsubscribes From Notifs

megandouli:
# what 
Updater crashes when trying to send notifications to a user who unsubscribed by replying "STOP".

# how 
Catch that specific error from Twilio and delete the user's subscriptions.


SimChadha (commented):
Looks good

SimChadha (approved):

PR #62, opened 2022-01-25
Add endpoint for posting feedback

hankewyczz:
# Purpose

The feedback/contact buttons do not work. This adds a feedback endpoint to the notifs server, which notifies the Search team on Slack.

<br>

# Tickets

https://trello.com/c/MzKg5kJT/288-feedback-buttons-are-inactionable

# Contributors

- @hankewyczz 

<br>

# Checklist

- [ ] Approved by designers
- [ ] Is passing linting checks
- [ ] Is passing tsc
- [ ] Is passing existing tests
- [ ] Has documentation and comments in code

<br>
@sandboxnu/searchneu

### DIFF
--- a/infrastructure/dev/docker-compose-server.yml
+++ b/infrastructure/dev/docker-compose-server.yml
@@ -36,5 +36,6 @@ services:
       TWILIO_VERIFY_SERVICE_ID:
       CLIENT_ORIGIN: http://localhost:5000
       JWT_SECRET:
+      SLACK_WEBHOOK_URL:
 volumes:
   pg:
--- a/twilio/server.ts
+++ b/twilio/server.ts
@@ -2,6 +2,7 @@ import express from "express";
 import cors from "cors";
 import { createServer } from "http";
 import jwt from "jsonwebtoken";
+import request from "request-promise-native";
 import twilioNotifyer from "./notifs";
 import notificationsManager from "../services/notificationsManager";
 import macros from "../utils/macros";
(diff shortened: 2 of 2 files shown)


hankewyczz:
When merging, the environment variable `SLACK_WEBHOOK_URL` must be added to the prod task definition in AWS. The variable is already in the Parameter Store.

andrewydai (changes requested):
I think this looks good codewise, but wondering if there is a need to make a request to CCA, and then to slack. Is there an issue with keeping the webhook on the front end env variables and issuing the request to slack from there? It would also stop us from adding an endpoint to our twilio server that doesn't seem relevant to notifs (I may be missing a security issue here?)

hankewyczz:
@andrewydai 

I'm split on this. My main reasoning for this was security, but it may be premature.

The Slack webhook has no authentication - the URL itself is the secret

> Keep it secret, keep it safe. Your webhook URL contains a secret. Don't share it online, including via public version control repositories. Slack actively searches out and revokes leaked secrets.
- https://api.slack.com/messaging/webhooks#getting_started


For that reason, I wasn't a fan of sending the request thru the frontend, as anyone who opens their console/network logs can see it. On the other hand, all they can do with the URL is spam the support channel - which they can do anyways. In either case, we can always revoke the webhook and create a new one, which is trivial. 

Basically, what it came down to for me was - if someone starts abusing it, we can add some rate-limiting controls on the backend. If someone is persistently spamming the channel (ie. using the new links as soon as we generate them), this would be a way of protecting against that. To send the request from the frontend, we'd have to expose the URL, which isn't great (I'm not really worried about people abusing this, but more likely a scraper finding it). 

I agree with you on the server issue - I don't like using the notifs server for this, but at the same time, spinning up another server for this would be overkill.

megandouli:
> @andrewydai
> 
> I'm split on this. My main reasoning for this was security, but it may be premature.
> 
> The Slack webhook has no authentication - the URL itself is the secret
> 
> > Keep it secret, keep it safe. Your webhook URL contains a secret. Don't share it online, including via public version control repositories. Slack actively searches out and revokes leaked secrets.
> 
> * https://api.slack.com/messaging/webhooks#getting_started
> 
> For that reason, I wasn't a fan of sending the request thru the frontend, as anyone who opens their console/network logs can see it. On the other hand, all they can do with the URL is spam the support channel - which they can do anyways. In either case, we can always revoke the webhook and create a new one, which is trivial.
> 
> Basically, what it came down to for me was - if someone starts abusing it, we can add some rate-limiting controls on the backend. If someone is persistently spamming the channel (ie. using the new links as soon as we generate them), this would be a way of protecting against that. To send the request from the frontend, we'd have to expose the URL, which isn't great (I'm not really worried about people abusing this, but more likely a scraper finding it).
> 
> I agree with you on the server issue - I don't like using the notifs server for this, but at the same time, spinning up another server for this would be overkill.

@hankewyczz @andrewydai FWIW last winter SearchNEU had a marathon to split the GraphQL server (aka course-catalog-api) from the notifications server. Unfortunately, when we switched to SMS notifs, we reverted the split and once again coupled CCA with the SearchNEU backend.

megandouli (approved):
just want to note that deploying this change will be non-trivial because we have a new env variable so we'll need to use Terraform @andrewydai happy to pair on this sometime if you'd like

andrewydai (approved):

PR #64, opened 2022-01-27
Ensure all tests work

hankewyczz:
Baby PR laying groundwork for the Prisma upgrade

- Updates `dbtests`, so I can actually tell if I break something in the upgrade
- Add tests as PR checks for GitHub
- Updates Node to `v14` (needed for the upgrade) - previously blocked by `pg`
- Updated `pg` and `jest` (hence large lockfile changes)


megandouli (commented):
does the `pg` upgrade require non-dev environments to also need a node upgrade? If so, we should also change the Dockerfile to start with an image of node 14 instead of node 12

hankewyczz:
good catch, thanks

tests/db_test_env.ts
```diff
@@ -36,8 +36,16 @@ class PrismaTestEnvironment extends NodeEnvironment {
     const client = new Client({
       connectionString: this.databaseUrl,
     });
-    await client.connect();
-    await client.query(`DROP SCHEMA IF EXISTS "${this.schema}" CASCADE`);
+    await client
+      .connect()
+      .then(() => console.log("connected"))
```
megandouli, line 41: nit: a better log message would be helpful in case a future dev wants to know where the log is coming from and greps for it in the codebase - maybe "connected to test database"

tests/db_test_env.ts
```diff
@@ -36,8 +36,16 @@ class PrismaTestEnvironment extends NodeEnvironment {
...
+      .catch((err) => console.log(err));
+    await client
+      .query(`DROP SCHEMA IF EXISTS "${this.schema}" CASCADE`)
+      .then(() => console.log("executed"))
```
megandouli, line 45: same thing here maybe log "successfully dropped schema if exists"

megandouli (approved):
awesome work!!

andrewydai (approved):
agree with Megan's comments but otherwise LGTM, nice job :)

hankewyczz:
Oh, those comments were actually for debugging and I kinda forgot about them. The `pg` issue was connecting to the database, executing the query, and dying without logging anything in postgres, so I was trying to narrow the issue down.

PR #65, opened 2022-01-28
Prisma Upgrade

hankewyczz:
Current Prisma version: `2.27.0`

Prisma uses semver, so the only breaking changes we need worry about are in `3.0.1`
https://github.com/prisma/prisma/releases/tag/3.0.1

Breaking Changes:
* :white_check_mark:  *Named Constraints*
    * Mitigated?
        * *NO*
    * Mitigation details:
        * Introspect the dev database to populate the Prisma schema with constraint and index names in your underlying database that do not match Prisma's naming convention:
    * Further reading
        * https://www.prisma.io/docs/guides/upgrade-guides/upgrading-versions/upgrading-to-prisma-3/named-constraints

* :white_check_mark:  *Referential Actions*
    * *Mitigated?*
        * *NO*
    * *Mitigation details:*
        * [Introspect ](https://www.prisma.io/docs/concepts/components/introspection)the database to populate the Prisma schema with a data model that reflects the current database schema.
        * Check all `onUpdate` and `onDelete` declarations, and ensure that they are all appropriate
    * Further reading:
        * https://www.prisma.io/docs/concepts/components/prisma-schema/relations/referential-actions#using-introspection

* :white_check_mark:  *`$queryRaw` API changes*
    * *Mitigated?*
        * *Yes*
    * *Mitigation details:*
        * There are multiple uses of `queryRaw` and `executeRaw`.  These uses cannot be converted to parameterized queries, since the column names are variable
        * We have to transition to `queryRawUnsafe` and `executeRawUnsafe`
    * Further Reading:
        * https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access#queryraw

* :white_check_mark: *The minimum required version of Node.js is v12.6*
    * *Mitigated?*
        * *YES*
    * *Mitigation details:*
        * Specified `nvmrc` to `14.18.3`, and changed `package.json` target to match

> *Not Applicable*

*  *:grey_question: Seeding*
    * *Mitigated?*
        * *Not applicable*
    * *Mitigation details:*
        * *We don't do any seeding*

* :grey_question:*Changes to how you query Null values on JSON fields*
    * *Mitigated?*
        * *Not applicable*
    * *Mitigation details:*
        * *There are no instances of filtering on a JSON field, let alone by `null`*

* :grey_question: *Renamed Aggregated Fields*
    * *Mitigated?*
        * *Not applicable*
    * *Mitigation details:*
        * *No aggregate fields are being used*


prisma/schema.prisma
```diff
@@ -9,127 +8,127 @@ datasource db {
 }
 
 model Course {
-  classAttributes String[]  @map(name: "class_attributes")
-  classId         String?   @map(name: "class_id")
+  classAttributes String[]         @map("class_attributes")
```
hankewyczz, line 11: Sorry for the style changes, but those are a byproduct of Prisma introspection. Since Prisma seems to prefer this format, it makes sense to keep it (even though it makes reviewing this PR a little harder)

prisma/schema.prisma
```diff
@@ -9,127 +8,127 @@ datasource db {
...
 
-  FollowedCourse FollowedCourse[]
-  @@map(name: "users")
+  @@map("users")
 }
 
 model TermInfo {
-  termId     String @unique @map(name: "term_id")
-  subCollege String @map(name: "sub_college")
+  termId     String @unique(map: "term_ids.term_id_unique") @map("term_id")
```
megandouli, line 129: was this change autogenerated by prisma? I don't get why the name mapping changed

megandouli (commented):
thanks for investigating this!! love the detail in the PR description

are you able to work with @andrewydai to deploy this change? seems like it'll be non-trivial. And if you know more detailed steps for deploying (e.g. what commands to run on the jumphost) could you put the steps in this PR description under a section called "Deployment Notes"?

puneetjohal (approved):
Other than what Megan already mentioned, looks good to me.

andrewydai (approved):
LGTM. One small thing, wondering if the `updater.ts` change from `queryRaw` to `queryRawUnsafe` is worth reverting and not using a template string here. Docs seem to indicate a risk of SQL injection through the template string, though I don't think that's an actual problem since this isn't user facing but just our updater service. Consequently I didn't let it stop my approval of the PR but just a passing thought

prisma/schema.prisma
hankewyczz: Yeah, it's Prisma. The way they name constraints was changed in v3, giving us two options:
- Use introspection, and let Prisma explicitly set the names in our `schema.prisma` to match the current naming in the database
- Allow Prisma to rename the properties to match the new defaults

I chose the first option, because this doesn't change our setup at all - it just makes our naming explicit. This makes the deployment trivial - we don't have to run any kind of upgrade/migration path, since everything's the same

Further reading: https://www.prisma.io/docs/guides/upgrade-guides/upgrading-versions/upgrading-to-prisma-3/named-constraints

hankewyczz:
@andrewydai 

The change from `queryRaw` to `queryRawUnsafe` isn't a functional change - it's just a naming change

> We’ve taken this opportunity to cleanup $queryRaw and $executeRaw. In Prisma 2.x, the Prisma Client had different behavior depending on how you called $queryRaw or $executeRaw:
*Tagged Template*: `prisma.$queryRaw...''`. This API sent a prepared statement and was safe from SQL injections.
*Function Call*: `prisma.$queryRaw(...)`. This API sent a raw query string and was not safe from SQL injections.

We were using the function call syntax, which has been renamed `queryRawUnsafe`.

Unfortunately, we're stuck with this, since the `unsafe` methods are the only way of passing a table or column name as a variable into a parameterized query.

PR #67, opened 2022-02-03
add initializing indices if not existing

andrewydai:
- Fixes an issue with the updater and elastic client where initial setup of indices was not occurring. Now, any bulk upsert will first check the existence of any blue/green index, and create them when necessary.
- Abstracts the naming and mapping of the index away from all business logic on elastic.ts so it only resides in the constructor (in case we need a future edit to indices as a whole)
-  Cleans up some more of `elastic.ts`

### DIFF
--- a/utils/elastic.ts
+++ b/utils/elastic.ts
@@ -63,19 +63,19 @@ export class Elastic {
   // This method fetches the exact index name since they are now dynamically named green or blue
   async fetchIndexName(aliasName: string): Promise<void> {
     const { mapping } = this.indexes[aliasName];
-    const isBlue = await this.doesIndexExist(`${aliasName}_blue`);
-    const isGreen = await this.doesIndexExist(`${aliasName}_green`);
+    const indexNames = [`${aliasName}_blue`, `${aliasName}_green`];
+
+    for (const indexName of indexNames) {
+      if (await this.doesIndexExist(indexName)) {
+        this.indexes[aliasName].name = indexName;
+        return;
+      }
+    }
 
     // if neither index exists, create a new index
-    if (!isBlue && !isGreen) {
-      const indexName = `${aliasName}_blue`;
-      await this.createIndex(indexName, mapping);
-      await this.createAlias(indexName, aliasName);
-    } else if (isBlue) {
-      this.indexes[aliasName].name = `${aliasName}_blue`;
-    } else {
-      this.indexes[aliasName].name = `${aliasName}_green`;
-    }
+    const indexName = indexNames[0];
+    await this.createIndex(indexName, mapping);
+    await this.createAlias(indexName, aliasName);
   }
 
   async isConnected(): Promise<boolean> {
@@ -93,7 +93,7 @@ export class Elastic {
       await client.indices.create({ index: indexName, body: mapping });
       macros.log(`Created index ${indexName}`);
     } catch (e) {
-      macros.log(`Error creating index ${indexName}: ${e}`);
+      macros.error(`Error creating index ${indexName}: ${e}`);
       throw e;
     }
   }
@@ -104,7 +104,7 @@ export class Elastic {
       await client.indices.delete({ index: indexName });
       macros.log(`Deleted index ${indexName}`);
     } catch (e) {
-      macros.log(`Error deleting index ${indexName}: ${e}`);
+      macros.error(`Error deleting index ${indexName}: ${e}`);
       throw e;
     }
   }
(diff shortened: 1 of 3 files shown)


utils/elastic.ts
```diff
@@ -78,44 +87,68 @@ export class Elastic {
     return true;
   }
 
+  async createIndex(indexName: string, mapping): Promise<void> {
+    macros.log(`Creating index ${indexName}`);
+    try {
+      await client.indices.create({ index: indexName, body: mapping });
+      macros.log(`Created index ${indexName}`);
+    } catch (e) {
+      macros.log(`Error creating index ${indexName}: ${e}`);
```
hankewyczz, line 96:
```suggestion
      macros.error(`Error creating index ${indexName}: ${e}`);
```

utils/elastic.ts
```diff
@@ -78,44 +87,68 @@ export class Elastic {
...
+    }
+  }
+
+  async createAlias(indexName: string, aliasName: string): Promise<void> {
+    macros.log(`Aliasing index ${indexName} as ${aliasName}`);
+    try {
+      await client.indices.putAlias({ index: indexName, name: aliasName });
+      macros.log(`Aliased index ${indexName} as ${aliasName}`);
+    } catch (e) {
+      macros.log(`Error aliasing ${indexName} as ${aliasName}: ${e}`);
```
hankewyczz, line 118:
```suggestion
      macros.error(`Error aliasing ${indexName} as ${aliasName}: ${e}`);
```

utils/elastic.ts
```diff
@@ -27,46 +27,55 @@ const BULKSIZE = 5000;
...
+    // if neither index exists, create a new index
+    if (!isBlue && !isGreen) {
+      const indexName = `${aliasName}_blue`;
+      await this.createIndex(indexName, mapping);
+      await this.createAlias(indexName, aliasName);
+    } else if (isBlue) {
+      this.indexes[aliasName].name = `${aliasName}_blue`;
+    } else {
+      this.indexes[aliasName].name = `${aliasName}_green`;
+    }
```
hankewyczz, line 78: [nit] I'm not a fan of way that in `doesIndexExist(NAME)` and `this.indexes[aliasName].name = NAME`, `NAME` is not a single variable. This could lead to typos down the line (but to be fair, I doubt we'll touch this often, which is why this is just a nit)

```suggestion
    const indexNames = [`${aliasName}_blue`, `${aliasName}_green`];
    
    for (const indexName of indexNames) {
      if (await this.doesIndexExist(indexName)) {
        this.indexes[aliasName].name = indexName;
        return;
      }
    }
    
    // if neither index exists, create a new index
    const indexName = indexNames[0];
    await this.createIndex(indexName, mapping);
    await this.createAlias(indexName, aliasName);
```

hankewyczz (commented):
Code looks good, I'll try the setup locally.

utils/elastic.ts
andrewydai: I like this, I also had the same thought but couldn't think of a reasonable solution at the time

megandouli (commented):
one small note, since you're moving the setup instructions to here, can you remove it from the CCA README?

puneetjohal (commented):
LGTM, and tested out these changes with you so should be good to go.

utils/elastic.ts
```diff
@@ -27,46 +27,55 @@ const BULKSIZE = 5000;
...
+    this.indexes = {};
+    this.indexes[this.CLASS_ALIAS] = classIndex;
+    this.indexes[this.EMPLOYEE_ALIAS] = employeeIndex;
   }
 
-  // This method fetches the exact index name since they are now dynamically named green or blue
-  async fetchIndexNames() {
-    const isClassesBlue = await this.doesIndexExist("classes_blue");
-    const employeesBlue = await this.doesIndexExist("employees_blue");
+  async fetchIndexNames(): Promise<void> {
```
SimChadha, line 56: Does the old purpose documentation no longer apply/ what would the new statement be?

utils/elastic.ts
```diff
@@ -27,46 +27,55 @@ const BULKSIZE = 5000;
...
+    const aliases = Object.keys(this.indexes);
+    await Promise.all(
+      aliases.map(async (alias) => await this.fetchIndexName(alias))
+    );
+  }
 
-    const classesName = isClassesBlue ? "classes_blue" : "classes_green";
-    const employeesName = employeesBlue ? "employees_blue" : "employees_green";
+  // This method fetches the exact index name since they are now dynamically named green or blue
```
SimChadha, line 63: What is does being named green/blue do in ES?

SimChadha (commented):
Code is readable/easy to understand and docs look good

utils/elastic.ts
andrewydai: Green blue is a common deployment paradigm for reducing downtime for services. To summarize, it involves creating an alias for the actual index instances so that we can operate on one before switching which one the alias points to. For example, we may have our alias point to the "blue" index, clone the index instance to a new index called "green", do some stuff on green, then switch the alias to point to green.

puneetjohal (approved):

PR #84, opened 2022-02-21
Up Prisma Connection limit

hankewyczz:
# Purpose

Bump the Prisma connection limit. 

Also did some looking into two other issues:
- Updater dying in prod
  - This has two causes - the Prisma issue (which this should fix), and a `429` error. The latter is a consequence of our Elasticsearch instance size. I'm fine with that one - the task gets recreated anyways, so I think we can live with this (since it lets us keep the free tier) 
- Old courses not being deleted
  - See `ESLG 0080` for Spring 2022 - added some logging to help me track down the issue 
<br>

# Tickets

-https://trello.com/c/Kx93LHkv/276-fix-updater-dying-in-prod

# Contributors

-

# Feature List

-
-
-

# Notes

<br>

# Checklist

- [ ] Approved by designers
- [x] Is passing tsc
- [x] Is passing existing tests

<br>
@sandboxnu/searchneu


megandouli (approved):
love bite sized PRs

megandouli (commented):
looks like database test is failing?

andrewydai (approved):

hankewyczz:
DB tests were failing (this is why i <3 workflow tests)

Added env interpolation, forced Prisma upgrade (`3.9.0` has an environment variable issue, so we need `^3.9.1`, and removed the `feature-preview` flag (which I forgot to remove earlier)

megandouli (approved):
cool @andrewydai we probably need to add a new environment variable

hankewyczz:
@megandouli I wasn't sure about that - since the `prisma/.env` file is tracked by Git, will its variables be used, or does a new one still have to be inserted into Parameter Store?

If we need to add a new one, Parameter Store doesn't allow env variable interpolation, so we might have to add the connection limit to the variable itself

megandouli:
> @megandouli I wasn't sure about that - since the `prisma/.env` file is tracked by Git, will its variables be used, or does a new one still have to be inserted into Parameter Store?
> 
> If we need to add a new one, Parameter Store doesn't allow env variable interpolation, so we might have to add the connection limit to the variable itself

oh hm you're right we might not need a new variable in the parameter store actually if the Prisma client uses the environment variable `DATABASE_URL` and does string interpolation to create the new environment variable

PR #85, opened 2022-02-22
Fix Elasticsearch employee insertion

hankewyczz:
# Purpose
- Fixes insertion of employees (and by extension, professor search cards)
- Adds tests to ensure this behavior doesn't regress. 
- Updates employee scraping source (and tests)

The new employee insertion mechanism deletes all old professor entries, and inserts the new ones. 

https://github.com/sandboxnu/searchneu/issues/146
<br>

# Tickets

- https://trello.com/c/HmL8jG6w/91-professor-card-missing

# Contributors

- @hankewyczz 
- @dillonhammer 

# Notes

<br>

# Checklist

- [x] Approved by designers
- [x] Is passing tsc
- [x] Is passing existing tests


andrewydai (approved):
Did you intend to comment out all the tests? Code change looks good

hankewyczz:
Yep, I started writing the tests, then realized that they're useless until employee scraping is fixed.

hankewyczz:
Coveralls only reports the coverage change from the first commit, which is kinda sucky, but the coverage went up

scrapers/employees/employees.ts
```diff
@@ -57,342 +51,88 @@ const request = new Request("Employees");
...
+  parseLettersResponse(response: EmployeeRequestResponse[]): EmployeeWithId[] {
+    this.people = response.map((employee) => {
+      const employee_obj = {
+        name: `${employee.FirstName} ${employee.LastName}`,
+        firstName: employee.FirstName,
+        lastName: employee.LastName,
+        primaryDepartment: employee.Department,
+        primaryRole: employee.PositionTitle,
+        phone: employee.PhoneNumber,
+        emails: [employee.Email],
```
andrewydai, line 116: I think we should change the professor table with a migration to not have emails if it's not included in the response from the new scraping endpoint and we're blowing away old data too

scrapers/employees/employees.ts
```diff
@@ -57,342 +51,88 @@ const request = new Request("Employees");
...
+    return this.people;
   }
 
-  async get(lastNameStart: string): Promise<Promise<void>> {
-    const jsessionCookie = await this.getCookiePromise();
+  async get(lastNameStart: string): Promise<EmployeeWithId[]> {
+    const xcsrfToken = await this.getCsrfToken();
 
-    macros.verbose("neu employee got cookie", jsessionCookie);
+    macros.verbose("neu employee got x-csrf token", xcsrfToken);
```
andrewydai, line 131: NIT but could capitalize the neu since we did it earlier in the `getCsrfToken` method

scrapers/employees/employees.ts
```diff
@@ -57,342 +51,88 @@ const request = new Request("Employees");
...
+      return { ...employee_obj, id: this.generateEmployeeId(employee_obj) };
     });
+
+    return this.people;
   }
 
-  async get(lastNameStart: string): Promise<Promise<void>> {
-    const jsessionCookie = await this.getCookiePromise();
+  async get(lastNameStart: string): Promise<EmployeeWithId[]> {
+    const xcsrfToken = await this.getCsrfToken();
```
andrewydai, line 129: I'm guessing the token doesn't expire while scraping since there isn't much logic to handle that. I'm wondering if it might be easier to just do `this.csrftoken` if we already instantiate it in main so that it's cached. I guess I see arguments for both tho as using the method feels safer even though I'm not sure it is (csrftoken doesn't get set to null again and there's no catch for if the response for scraping or csrftoken fails).

andrewydai (changes requested):
Looks great! I left some comments. Also I know you didn't do type it as `any` for `processProf` but since we have the type `EmployeeWithId[]` it should be the type for profDump and we can maybe type its argument with it for a quick type check win

scrapers/employees/employees.ts
hankewyczz: Some of the department-specific webpages actually have multiple emails, so this should be kept at the moment. 

That does bring up a good point about department-specific scraping - why bother? Take `bigPictureUrl` for example - only the CCIS site actually has this attribute. We don't use it for the employee cards, so we're scraping it and not using it. 

There might be a reason for this that I'm unaware of, but I'm in favor of not using the department-specific scrapers. It's more code to maintain, complicates our process, and doesn't add much value as far as I can tell.

andrewydai: Fair point, might be worth ditching them if we can get all data necessary for employees from this new api. Also i'm assuming that all the employees NEU has are listed here, and we won't miss any employees by not scraping departments (that would be my main concern but i'd be surprised if it was true)

scrapers/employees/employees.ts
```diff
@@ -3,46 +3,40 @@
...
+import { v4 as uuidv4 } from "uuid";
 
 const request = new Request("Employees");
 
-// Scrapes from here: https://prod-web.neu.edu/wasapp/employeelookup/public/main.action
+function employeeQuery(query: string): unknown {
+  return {
+    versionInfo: {
+      moduleVersion: "HkL1qOMsAyyecvGQd_lOfQ",
+      apiVersion: "ad_141F5PYcK3c+D5K8O+g",
```
megandouli, line 22: nit: can we add comments here explaining where we got these values from?

megandouli (approved):

andrewydai (approved):

PR #95, opened 2022-03-10
add phrase searching to searcher

sebwittr:
# Purpose

Added the capabilities for phrase searching to the project. If a query has a phrase within quotes, the searcher will search using a match_phrase query. Every result must contain the contents of a phrase. 

To replicate locally, run yarn dev on the backend, and yarn dev:fullstack on the front end. You can search data structures, and see that discrete structures comes before it. If you search "data structures", it'll be the only result on the page. 
<br> 

# Tickets

- https://trello.com/c/lYrJEVja/269-add-phrase-searching

# Contributors

- @sebwittr 
- @andrewydai 

# Feature List

- Added a parser function to the searcher, which broke down queries to phrase queries and non-phrase queries
- Created a QueryWithType type, which was either "phrase" or "most_fields".
- New query structure requires result to include phrase, and should have any other terms.

# Notes

<br>

# Checklist

- [ ] Approved by designers
- [ ] Is passing tsc
- [ ] Is passing existing tests

<br>
@sandboxnu/searchneu

### DIFF
--- a/services/searcher.ts
+++ b/services/searcher.ts
@@ -292,11 +292,6 @@ class Searcher {
       fieldQuery = matchQueries[1][0];
     }
 
-    //if there was an empty search, match_all
-    if (phraseQueries.length === 0 && !fieldQExists) {
-      phraseQueries.push(MATCH_ALL_QUERY);
-    }
-
     // use lower classId has tiebreaker after relevance
     const sortByClassId: SortInfo = {
       "class.classId.keyword": { order: "asc", unmapped_type: "keyword" },
@@ -335,7 +330,10 @@ class Searcher {
       sort: ["_score", sortByClassId],
       query: {
         bool: {
-          must: phraseQueries, //must match all the phrases
+          must:
+            phraseQueries.length === 0 && !fieldQExists
+              ? MATCH_ALL_QUERY
+              : phraseQueries,
           filter: {
             bool: {
               should: [
--- a/services/searcher.ts
+++ b/services/searcher.ts
@@ -34,7 +34,7 @@ import {
   AggResults,
   SearchResult,
   CourseSearchResult,
-  QueryWithType,
+  ParsedQuery,
 } from "../types/searchTypes";
 
 type CourseWithSections = Course & { sections: Section[] };
@@ -220,50 +220,54 @@ class Searcher {
    * Most_field queries are everything else, and search the terms on all fields.
    * else.
    * @param query the string that is parsed into a list
-   * @returns a list containing the list of phrase queries and list of containing the most field query.
+   * @returns an object containing a list of phrase_queries, and a field query.
    */
-  parseQuery(query: string): LeafQuery[][] {
-    const queries = [];
+  parseQuery(query: string): ParsedQuery {
+    const queries = {};
 
     let matches = query.match(/"(.*?)"/gi);
-
     //if there are no phrases, then make matches an empty list instead of being null
     if (matches === null) {
       matches = [];
     }
 
     const re = new RegExp(matches.join("|"), "gi");
-    const non_matches = query.replace(re, "").replace('"', "");
 
+    //make sure theres no extra white space after removing the phrases.
+    const non_matches = query
+      .replace(re, "")
+      .replace('"', "")
+      .replace(/\s+/g, " ")
+      .trim();
     const phraseQueries = [];
 
     // go through the phrases, and make phrase queries with them.
     for (let i = 0; i < matches.length; ++i) {
       phraseQueries.push({
         multi_match: {
-          query: matches[i],
+          query: matches[i].substring(1, matches[i].length - 1),
           type: "phrase",
           fields: this.getFields(),
         },
       });
     }
 
-    queries.push(phraseQueries);
-
     //make the field query and add it to the list.
-    const fieldQuery = [];
+    let fieldQuery: LeafQuery = null;
     if (non_matches.trim() !== "") {
-      fieldQuery.push({
+      fieldQuery = {
         multi_match: {
           query: non_matches,
           type: "most_fields",
           fields: this.getFields(),
         },
-      });
+      };
     }
 
-    queries.push(fieldQuery);
-    return queries;
+    return {
+      phraseQ: phraseQueries,
+      fieldQ: fieldQuery,
+    };
   }
 
   /**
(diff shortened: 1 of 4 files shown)


services/searcher.ts
```diff
@@ -302,7 +386,7 @@ class Searcher {
     const queries: EsQuery[] = [
       this.generateQuery(query, termId, validFilters, min, max),
     ];
-
+    console.log(JSON.stringify(queries[0]));
```
megandouli, line 389: remove?

services/searcher.ts
```diff
@@ -213,6 +214,49 @@ class Searcher {
     ];
   }
 
+  /**
+   * Given a string, creates a list of queries that are either phrase queries or
+   * most_field queries. Phrases are between quotes, and use phrase matching.
+   * Most_field queries are everything else, and search the terms on all fields.
+   * else.
```
megandouli, line 221: did you mean to include `else.`?

services/searcher.ts
```diff
@@ -213,6 +214,49 @@ class Searcher {
     ];
   }
 
+  /**
+   * Given a string, creates a list of queries that are either phrase queries or
+   * most_field queries. Phrases are between quotes, and use phrase matching.
+   * Most_field queries are everything else, and search the terms on all fields.
```
megandouli, line 220: nice comment!

services/searcher.ts
```diff
@@ -213,6 +214,49 @@ class Searcher {
...
+   * else.
+   * @param query the string that is parsed into a list
+   * @returns a list of queries
+   */
+  parseQuery(query: string): QueryWithType[] {
+    const ret = [];
```
megandouli, line 226: nit: rename `ret` to `queries`

services/searcher.ts
```diff
@@ -213,6 +214,49 @@ class Searcher {
...
 
+  /**
+   * Given a string, creates a list of queries that are either phrase queries or
+   * most_field queries. Phrases are between quotes, and use phrase matching.
+   * Most_field queries are everything else, and search the terms on all fields.
+   * else.
+   * @param query the string that is parsed into a list
+   * @returns a list of queries
+   */
+  parseQuery(query: string): QueryWithType[] {
```
megandouli, line 225: let's add a unit test for `parseQuery` so we can ensure the quote stuff is correct

services/searcher.ts
```diff
@@ -213,6 +214,49 @@ class Searcher {
...
+        curString += query[i];
+      }
+    }
+    //if the current string is not empty, add whatever we have left.
+    if (curString !== "") {
+      ret.push({
+        query: curString,
+        type: isPhrase ? "phrase" : "most_fields",
+      });
+    }
```
megandouli, line 255: instead of all this logic examining character by character, I think we can use regex matching to extract phrases inside quotes - this regex should work: `"/(.*?)"/g` - I grabbed it from https://www.regextester.com/97161

services/searcher.ts
```diff
@@ -225,17 +269,55 @@ class Searcher {
...
+
+    //a most_field query, containing all non-phrase terms combined.
+    let fieldQuery: LeafQuery;
+
+    //are there any field Queries?
+    let fieldQExists = false;
+
+    //the current fields string, we want to build a single query with all
+    //the terms that are not phrases.
+    let fieldStr = "";
```
megandouli, line 286: I think it'd be cleaner to make `fieldStr` an array of strings and do a `.join(" ")` at the end. Seems better than having to manually concatenate strings with a space everytime

services/searcher.ts
```diff
@@ -225,17 +269,55 @@ class Searcher {
...
+    for (let i = 0; i < queryArr.length; ++i) {
```
megandouli, line 287: we don't actually need the index here so it'd more idiomatic to do `queryArr.forEach(element => {...})`

megandouli (dismissed):
cool feature! minor javascript things that could be improved

services/searcher.ts
hankewyczz: Piggy-backing off of this - I can't think of a straightforward way of matching the obvious. 

So, to get the non-quoted queries:
```
const matches = str.match(/"(.*?)"/gi);
const re = new RegExp(matches.join('|'), 'gi');
const non_matches = str.replace(re, '');
```
Note that this will leave the `"`'s behind, so we should probably strip them out as well.

hankewyczz (changes requested):
Generally looks good - just Megan's comments (I didn't see anything needing tweaks besides those)

services/searcher.ts
```diff
@@ -213,6 +214,60 @@ class Searcher {
...
+   * else.
+   * @param query the string that is parsed into a list
+   * @returns an object containing a list of phrase_queries, and a field query.
+   */
+  parseQuery(query: string): ParsedQuery {
+    let matches = query.match(/"(.*?)"/gi);
+    //if there are no phrases, then make matches an empty list instead of being null
+    if (matches === null) {
+      matches = [];
+    }
```
hankewyczz, line 230:
```suggestion
    let matches = query.matchAll(/"(.*?)"/gi) || [];
```

services/searcher.ts
hankewyczz: Was marked as resolved but the comment's still there

services/searcher.ts
```diff
@@ -213,6 +214,60 @@ class Searcher {
...
+
+    const re = new RegExp(matches.join("|"), "gi");
+
+    //make sure theres no extra white space after removing the phrases.
+    const non_matches = query
```
hankewyczz, line 235:
```suggestion
    const nonMatches = query
```
Case consistency

services/searcher.ts
```diff
@@ -213,6 +214,60 @@ class Searcher {
...
+    // go through the phrases, and make phrase queries with them.
+    for (let i = 0; i < matches.length; ++i) {
+      phraseQueries.push({
+        multi_match: {
+          query: matches[i].substring(1, matches[i].length - 1),
+          type: "phrase",
+          fields: this.getFields(),
+        },
+      });
+    }
```
hankewyczz, line 251:
```suggestion
    const phraseQueries = matches.map((match) => {
        return {
          multi_match: {
            query: match[1],
            type: "phrase",
            fields: this.getFields(),
          },
        }
    });
```

The way `matchAll` works: for each match, it returns an array of matches. 

Each match contains, in order:
- The full match
- Any capturing groups (which on our case, is everything but the quotes)


Make sure to test this, since I'm just writing this code in github and haven't 100% checked it

hankewyczz (changes requested):
Just a few more tweaks

services/searcher.ts
```diff
@@ -213,6 +214,60 @@ class Searcher {
...
+   * @returns an object containing a list of phrase_queries, and a field query.
+   */
+  parseQuery(query: string): ParsedQuery {
+    let matches = query.match(/"(.*?)"/gi);
+    //if there are no phrases, then make matches an empty list instead of being null
+    if (matches === null) {
+      matches = [];
+    }
+
+    const re = new RegExp(matches.join("|"), "gi");
```
megandouli, line 232: nit: rename `re` to `matchesRegExp`? `re` isn't very clear

services/searcher.ts
```diff
@@ -224,18 +279,13 @@ class Searcher {
...
-            },
-          }
-        : MATCH_ALL_QUERY;
+    //a list of all queries
+    const matchQueries: ParsedQuery = this.parseQuery(query);
+
+    const phraseQueries: LeafQuery[] = matchQueries.phraseQ;
+
+    const fieldQuery: LeafQuery = matchQueries.fieldQ;
+    const fieldQExists = fieldQuery !== null;
```
megandouli, line 288: nit: call this `fieldQueryExists` - shortening this variable name is not worth the cost of understandability on line 329

megandouli (commented):
nice! I'll leave it to @hankewyczz to approve but overall looks good to me (just a couple minor nits)

hankewyczz (approved):
🚀

andrewydai (approved):
👑

PR #97, opened 2022-03-18
Optimize employee scraping

hankewyczz:
Optimizations for employee scraping (now just 3 requests total!) and adding tests. 

Also, the NEU API returns strings of `Not Available` instead of `null` for empty parameters, which isn't what we want - adds handling for this.


scrapers/employees/employees.ts
```diff
@@ -15,113 +15,96 @@ import { v4 as uuidv4 } from "uuid";
...
+      })
+      .then((resp) => resp.body["versionToken"]);
+
+    const employeeQuery = {
+      versionInfo: {
+        moduleVersion,
+        apiVersion: "ad_141F5PYcK3c+D5K8O+g",
+      },
+      viewName: "MainFlow.FacultyAndStaffDirectory",
+      // Their frontend requires non-null queries, but the backend doesn't :)
```
andrewydai, line 89: to this point, do you think it's worth commenting out the code for the cartesian product of last names in case they start enforcing on backend?

tests/workflow_tests/scrape.test.git.ts
```diff
@@ -0,0 +1,17 @@
+import employees from "../../scrapers/employees/employees";
+
+describe("scraping employees", () => {
+  it("should be able to query the API", async () => {
+    await employees.queryEmployeesApi();
+
+    // It's a safe bet that there are at least 1000 employees
+    expect(employees.people.length).toBeGreaterThan(1_000);
```
andrewydai, line 8: If using cache maybe we can consider using an exact check bc it's static

andrewydai (changes requested):
I think it lgtm, would like more logging but unsure if you're covering that in other PR for scraping logging.

scrapers/employees/employees.ts
hankewyczz: I'd rather not keep the whole code (unnecessary clutter), but I added a comment explaining a bit (and linking this PR)

PR #118, opened 2022-05-27
Testing

hankewyczz:
I <3 testing!!!


scrapers/classes/processors/tests/addPrerequisiteFor.test.js
```diff
@@ -3,130 +3,233 @@
  * See the license file in the root folder for details.
  */
 
+import Keys from "../../../../utils/keys";
```
sebwittr, line 6: is this import necessary? Lint checker says Keys is never being used.

scrapers/employees/util.ts
```diff
@@ -133,6 +133,8 @@ export function getBaseHost(url: string): null | string {
     return null;
   }
 
+  console.log(homepage);
```
sebwittr, line 136: get rid of console.log

tests/end_to_end/updater.test.git.ts
```diff
@@ -26,6 +26,7 @@ describe("Searching for courses", () => {
         }
       }
     `);
+    console.log(JSON.stringify(res, null, 2));
```
sebwittr, line 29: was this console.log left in accidentally?

sebwittr (changes requested):
Everything looks really good! These tests definitely cover a lot, and a lot of the code is much cleaner. There are a couple minor linting changes that need to be fixed, but other than that I see no problems.

sebwittr (approved):
All the changes look good!

PR #138, opened 2022-10-21
Added section filtering for Honors and Campus

sebwittr:
# Purpose

<br> Filtering by honors and campus wouldn't actually filter out the wrong sections. This PR does further filtering allowing these sections to be filtered.

# Tickets

- https://trello.com/c/eG62qJDF/262-campus-section-filter

# Contributors

- @sebwittr @Lucas-Dunker 

# Feature List

- Further filters the Elastic Search result list, removing any sections that don't match the filters and any classes that may get filtered out altogether.
-
-

# Notes

<br>

# Checklist

<br>
@sandboxnu/searchneu


Lucas-Dunker (approved):
Looks good!

services/searcher.ts
```diff
@@ -482,17 +483,71 @@ class Searcher {
...
     const hydrateDuration: number = Date.now() - startHydrate;
 
+    // Do extra filtering if Honors or Campus Filters are being applied since those filter sections,
+    // which ElasticSearch doesn't actually filter out.
+
+    const validFilters = this.validateFilters(filters);
+    let honorsFilter = false;
+    let campusFilter = "";
+
+    // are we filtering by honors?
```
hankewyczz, line 502: Generally, we should omit comments like these - it's simple enough that the code explains itself. Comments should be reserved for areas of the code which are pretty complex/unintuitive.

services/searcher.ts
```diff
@@ -482,17 +483,71 @@ class Searcher {
...
+    if (Object.keys(validFilters).includes("honors")) {
+      honorsFilter = true;
+    }
```
hankewyczz, line 505:
```suggestion
    const honorsFilter = "honors" in validFilters;
```

services/searcher.ts
```diff
@@ -482,17 +483,71 @@ class Searcher {
...
+
+    // are we filtering by campus?
+    if (Object.keys(validFilters).includes("campus")) {
+      campusFilter = "" + validFilters["campus"];
+    }
```
hankewyczz, line 510:
```suggestion
    const campusFilter = validFilters["campus"] ?? "";
```

Same comment as above - try to remove this comment, since it doesn't add much. 

For context - the `??` is the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator). If the argument on the left is `null`, it returns the argument on the right. So, if `validFilters["campus"]` doesn't exist, it defaults to `""`

services/searcher.ts
```diff
@@ -482,17 +483,71 @@ class Searcher {
...
+
+    // create a new list initially containing the results, if we are doing further filtering
+    // we'll set it to an empty list and build up the new filtered results.
+    let filteredResults: SearchResult[] = results;
+    if (honorsFilter || campusFilter != "") {
```
hankewyczz, line 515:
```suggestion
    if (honorsFilter || campusFilter !== "") {
```

services/searcher.ts
```diff
@@ -482,17 +483,71 @@ class Searcher {
...
+
     const hydrateDuration: number = Date.now() - startHydrate;
 
+    // Do extra filtering if Honors or Campus Filters are being applied since those filter sections,
+    // which ElasticSearch doesn't actually filter out.
+
+    const validFilters = this.validateFilters(filters);
+    let honorsFilter = false;
+    let campusFilter = "";
+
```
hankewyczz, line 501: All of this should be moved into a new function. Our codebase doesn't do a very good job of it, but one function === one job. Also, a readability rule-of-thumb - if you can't fit the entire function on your screen at the same time, it's too big.

services/searcher.ts
```diff
@@ -482,17 +483,71 @@ class Searcher {
...
+          curResult.sections = curResult.sections.filter(
+            (curSection: Section) => {
+              if (honorsFilter && !curSection.honors) {
+                return false;
+              }
+              if (campusFilter !== "" && curSection.campus !== campusFilter) {
+                return false;
+              }
+              return true;
+            }
```
hankewyczz, line 531: I'd recommend breaking this logic into a seperate function as well, and calling:
```ts
filteredResults = results.filter(this.filterFunction(honorsFilter, campusFilter));
```

with an actual name for the filter function lol

services/searcher.ts
```diff
@@ -482,17 +483,71 @@ class Searcher {
...
+        }
+      });
+    }
+
     return {
-      searchContent: results,
-      resultCount,
+      searchContent: filteredResults,
+      // the result count might differ if some classes get completely filtered out, so we set it to the
+      // length of filteredResults.
```
hankewyczz, line 549: Personally, I also think this comment is unnecessary. 

As a general aside - if you feel differently, feel free to keep it. My main reasoning is that we now have two "sources of truth" for whats going on in the code. If the code is updated, and that person forgets to update the comment, we now have two different interpretations of what's going on, and we don't know which one to trust. And devs always forget to update comments eventually :( (eg. try reading `requests.ts`)

hankewyczz (changes requested):
Looking good, just some style/readability changes 💯

hankewyczz:
@Lucas-Dunker as an aside - we generally don't want to review our own PRs. Thinking about it - if you catch an error in the code, you'll just fix and commit it. And if you don't catch an error, having a green "Approved" on the PR might make other people think it's had a fresh set of eyes on it already.

services/searcher.ts
```diff
@@ -458,6 +459,60 @@ class Searcher {
...
+    if (campusFilter !== "" && curSection.campus !== campusFilter) {
+      return false;
+    }
+
+    return true;
+  }
+
+  filterResults(filters, results): SearchResult[] {
+    // Do extra filtering if Honors or Campus Filters are being applied since those filter sections,
+    // which ElasticSearch doesn't actually filter out.
```
hankewyczz, line 479:
```suggestion
  /** 
   * Perform extra filtering if Honors or Campus Filters are being applied.
   * Elasticsearch will return ALL sections for a class if ANY of its sections match the filter.
   * eg. if a class has 100 sections and 1 of them is Honors, applying the Honors filter
   *       in Elasticsearch will return all 100 sections.
   * This function only returns the sections which match the filter.
   */
  filterResults(filters, results): SearchResult[] {
```

For function comments, we prefer [JSDoc](https://jsdoc.app/) - that way, when someone hovers over the function name, they'll see the comment. (I know the majority of our codebase doesn't use that, but we want to write better code going forward & slowly refactor the old code to match)

services/searcher.ts
```diff
@@ -458,6 +459,60 @@ class Searcher {
...
+
+    const honorsFilter: boolean = validFilters["honors"] ? true : false;
+
+    const campusFilter: string = "" + (validFilters["campus"] ?? "");
+
+    // create a new list initially containing the results, if we are doing further filtering
+    // we'll set it to an empty list and build up the new filtered results.
+    let filteredResults: SearchResult[] = results;
+    if (honorsFilter || campusFilter !== "") {
+      filteredResults = [];
```
hankewyczz, line 491:
```suggestion
    if (!honorsFilter && campusFilter === "") {
      return results;
    }
    .... // handle the filtering here
```

Instead of wrapping the "we need to filter further" case in an `if` block, just write a simple `if` to see if we DON'T need to filter further. This is preferred because:
- we don't have to wrap a large chunk of code in a block - it's simpler & easier to read (and less indentation)
- instead of assigning `filteredResults` to `results` and conditionally changing that, just assign `const filteredResults = []`. It's simpler, doesn't require a comment explaining it, and allows you to use `const`

services/searcher.ts
```diff
@@ -458,6 +459,60 @@ class Searcher {
...
+
+      // go through each Result, if it's a course: filter the sections.
+      results.forEach((curResult: SearchResult) => {
+        if (curResult.type === "class") {
+          // filter the sections for each result
```
hankewyczz, line 496: Generally, we should omit comments like these - it's simple enough that the code explains itself. Comments should be reserved for areas of the code which are pretty complex/unintuitive.

I'm not going to leave a review comment on each spot like this - pls just go through and check all the comments to see if they generally meet the classification above.

services/searcher.ts
```diff
@@ -458,6 +459,60 @@ class Searcher {
...
+    return true;
+  }
+
+  filterResults(filters, results): SearchResult[] {
+    // Do extra filtering if Honors or Campus Filters are being applied since those filter sections,
+    // which ElasticSearch doesn't actually filter out.
+
+    const validFilters = this.validateFilters(filters);
+
+    const honorsFilter: boolean = validFilters["honors"] ? true : false;
```
hankewyczz, line 483:
```suggestion
    const honorsFilter = validFilters["honors"] ?? false;
```

We want to use the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) here.

Using a conditional in the format 
```js
if (condition) {
   return true
} else {
   return false
}
```
is generally a bad design - you could just return the `condition`. However, in this case, the conditional is necessary for Javascript to cast the type to a boolean. We want to avoid that, since that sort of casting can mask errors (eg. if for some reason `"honors": "error"`, the string `"error"` would cast to `true` (since it's non-empty). However, that should break our code, since its a completely unexpected value). 

Note - read the [Examples page for the `??` operator - it's quick](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing#examples)

services/searcher.ts
```diff
@@ -458,6 +459,60 @@ class Searcher {
...
+
+    const campusFilter: string = "" + (validFilters["campus"] ?? "");
```
hankewyczz, line 485:
```suggestion
    const campusFilter = validFilters["campus"] ?? "";
```

Same comment as above - we don't want to use the `"" + ...` pattern to forcibly convert `validFilters["campus"]` to a string, since that can mask errors. Also, the explicit type isn't necessary here - TS is good at figuring out basic types like this (you can leave it, ofc, just figured it's worth pointing out)

services/searcher.ts
```diff
@@ -458,6 +459,60 @@ class Searcher {
     };
   }
 
+  filterSection(
```
hankewyczz, line 462: Can we add a JSDoc comment for this function?

hankewyczz (changes requested):
a few more minor tweaks

as an aside, I'd appreciate it if both of you could go through this feedback together - hopefully, both of you might get something from these comments

hankewyczz (approved):
looks great!! 💯

hankewyczz:
@sebwittr When you get a chance, could you merge this branch with the updated master branch?

hankewyczz:
Fixed in #156

PR #139, opened 2022-10-25
Patch updater crashing for term ID `202225`

hankewyczz:
# Purpose

<br>
For some reason, the term with ID `202225` exists in a weird state in Banner:
- It is on the list of term IDs
- However, there are no sections associated with this term ID
  - As a result - when we call the Banner endpoint to get a term's sections, Banner throws an error
  - To mitigate this, we'll just filter this term out when updating. It's a past term anyways, so these classes/sections aren't changing anyways. 

# Checklist

- [ ] Has documentation and comments in code

<br>
@sandboxnu/searchneu

### DIFF
--- a/services/updater.ts
+++ b/services/updater.ts
@@ -17,6 +17,8 @@ import { NotificationInfo } from "../types/notifTypes";
 
 import { NUMBER_OF_TERMS_TO_UPDATE } from "../scrapers/classes/parsersxe/bannerv9Parser";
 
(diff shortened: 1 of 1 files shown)


tiingweii-shii (approved):
Classic Banner 🤖

Lucas-Dunker (commented):
Easy fix, I'd recommend digging further into the root of the problem but since this is on banners end we can only do so much 😔

Lucas-Dunker (approved):

PR #143, opened 2022-10-28
Mitigate Prisma connection pool timeouts

hankewyczz:
# Purpose

We're still having issues with Prisma, despite the upgrades. However, Prisma issues are no longer popping up in the webserver (the actual API). Now, it's only visible in the updater - the task that runs every 5 minutes. 

It appears that Prisma just doesn't handle parallel connections well (even though that's the purpose of the connection pool). Related issues are https://github.com/prisma/prisma/issues/13681 and https://github.com/prisma/prisma/issues/12858

For the time being, this ticket does the following:
- Upgrade to Prisma `4.x`
  - This doesn't fix our issues at the moment, but upgrading now will ease the upgrade path in the future (if the connection pool issues get fixed)
- Makes some of our calls serialized
  - They used be in parallel, but serializing mitigates the Prisma issues
  - Timing isn't a huge concern for us - the updater runs once every 5 minutes, and each run takes about 30 seconds, so we have plenty of time to spare  

<br>

# Tickets

- https://trello.com/c/triEJfl1/334-upgrade-to-prisma-4x

# Contributors

- @hankewyczz 

<br>

# Checklist

<br>
@sandboxnu/searchneu


infrastructure/terraform/modules/course-catalog-api/domain.tf
```diff
@@ -49,7 +49,7 @@ resource "aws_lb_target_group" "webserver" {
   health_check {
     matcher = "200"
     path = "/.well-known/apollo/server-health"
-    unhealthy_threshold = 3
+    unhealthy_threshold = 5
```
hankewyczz, line 52: This was causing issues in AWS - unsure exactly why.

Basically, the health check runs every 30s. If three in a row fail, the container is killed. That gives us 1m30s to successfully start up the container. 

However, with Prisma `4.x`, the setup for the container (ie. `migrate` and `refresh`) takes more than 2 minutes. With the old health checks, this would kill the container before it's even set up, and spin up a new container - which would have the same issue. 

Unfortunately, there is no way to create a "grace period" during which the health check doesn't run (as far as I could find)

package.json
```diff
@@ -19,14 +19,14 @@
     "dev:docker:del": "docker-compose -f infrastructure/dev/docker-compose.yml down -v",
     "dev:docker:server": "docker-compose -f infrastructure/dev/docker-compose-server.yml up -d",
     "dev:docker:server:del": "docker-compose -f infrastructure/dev/docker-compose-server.yml down -v",
-    "db:migrate": "npx prisma migrate dev",
-    "db:refresh": "npx prisma generate",
+    "db:migrate": "yarn prisma migrate dev",
```
hankewyczz, line 22: We no longer use `npx` - this just switches fully over to `yarn`

hankewyczz: This doesn't have an impact of the code/performance - it's just something I noticed that we need to change

sebwittr (approved):
Every change makes sense. Hopefully serialization and updating the health checks fix the problems Search has been having!

DilanRocco:
👍

PR #145, opened 2022-10-30
Added "isCurrentTerm" attribute to TermInfo

DilanRocco:
# Purpose

Added the functionality to determine if a specific term is the current term. 
For the website SearchNEU, this is important because currently users can enable notifications for open seats of prior semesters. By adding this, we can determine if a semester is current or not, and decide whether to give the user the option to enable notifications.

Note: Since we decided that isCurrentTerm default value should be true, the tests will include the value isCurrentTerm to be True. isCurrentTerm does not get updated until it is ran through the scraper.
<br>

# Tickets

-https://trello.com/c/UBYWrMvL/306-sections-in-past-semesters-should-not-allow-you-to-sign-up-for-notifs

# Contributors

-@dilanrocco
-@sebwittr

# Feature List

- TermInfo now has an attribute isCurrentTerm which is a number(time since 1970) that stores the the last day in the semester.
- Designed a function isCurrentTerm to determine if the given term is a current term.
 

<br>

# Checklist

<br>
@sandboxnu/searchneu


graphql/resolvers/search.ts
```diff
@@ -12,8 +14,47 @@ interface SearchResultItemConnection {
...
+  isCurrentTerm: boolean;
 }
+function determineIfCurrentTerm(
+  maxEndDate: number,
+  resultSearch: SearchResult[],
+  termId: string
+): boolean {
+  // (e.g today is 9/20/22 and the semester ends 8/30/22) => not a current term
+  const date = new Date().getTime();
+  const currentDate = Math.floor(date / 8.64e7);
```
hankewyczz, line 26: Is this variable really the current date? `Date.now` is also the current date; we want to be more precise in variable naming. Maybe `daysSinceEpoch`?

graphql/resolvers/search.ts
```diff
@@ -12,8 +14,47 @@ interface SearchResultItemConnection {
...
+  if (maxEndDate != null && maxEndDate != -1) {
```
hankewyczz, line 27: `maxEndDate` should never be null, so this check is unnecessary. The type of `maxEndDate` is `number`, which doesn't allow for null values. 

Also, why would `maxEndDate` ever equal `-1`? This is another instance of "magic values"

graphql/resolvers/search.ts
```diff
@@ -12,8 +14,47 @@ interface SearchResultItemConnection {
...
+        maxEndDate = result.sections[0].meetings[0].endDate;
+      }
+      //     result.sections != undefined &&
+      //     result.sections[0].meetings != undefined &&
+      //     result.sections[0].meetings[0].endDate != undefined &&
+      //     result.sections[0].meetings[0].endDate != null &&
+      //     result.sections[0].meetings[0].endDate > maxEndDate
+      //   ) {
+      //
+      //   }
```
hankewyczz, line 49: Read this page about the [optional chaining operator.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) Basically, it allows us to avoid exactly this scenario in a really compact and nice manner

```suggestion
      const endDate = result?.sections[0]?.meetings[0]?.endDate;
      if (endDate !== undefined && endDate > maxEndDate) {
        maxEndDate = endDate;
      }
```

graphql/resolvers/search.ts
```diff
@@ -12,8 +14,47 @@ interface SearchResultItemConnection {
...
+    }
+  }
+  prisma.termInfo.update({
```
hankewyczz, line 52: this is an asynchronous function call, so unless you `await`, you won't know when it actually executes

graphql/resolvers/search.ts
```diff
@@ -12,8 +14,47 @@ interface SearchResultItemConnection {
...
+    where: { termId: termId },
+    data: { maxEndDate },
+  });
```
hankewyczz, line 55: This `prisma.update` should be in a separate function. One job == one function. Since this function is called `determineIfCurrentTerm`, this shouldn't be a part of it

graphql/resolvers/search.ts
```diff
@@ -52,8 +93,20 @@ const resolvers = {
         )
       );
 
+      const termInfo: TermInfo = await prisma.termInfo.findFirst({
+        where: { termId: "" + args.termId },
```
hankewyczz, line 97: There's no need to type-cast here - we know the term ID is a string. 

As a rule of thumb - if you find yourself doing that in Typescript, that probably means your types are inaccurate/not specific enough. We should almost never have to do that

```suggestion
        where: { termId: args.termId },
```

graphql/resolvers/search.ts
```diff
@@ -52,8 +93,20 @@ const resolvers = {
...
+      });
+      let maxEndDate = -1;
```
hankewyczz, line 99: This should throw an error. You've just executed a search, so you know that the term ID exists. If it doesn't exist for some reason, that should immediately throw an error.

scrapers/classes/parsersxe/termListParser.ts
```diff
@@ -24,6 +24,7 @@ class TermListParser {
         termId: term.code,
         text: text,
         subCollege: subCollege,
+        maxEndDate: -1,
```
hankewyczz, line 27: Since `maxEndDate` is an optional field, we shouldn't initialize it unless we have real data for it

scrapers/classes/parsersxe/tests/__snapshots__/termListParser.test.js.snap
```diff
@@ -4,18 +4,21 @@ exports[`termListParser pulls out relevant data 1`] = `
 Array [
   Object {
     "host": "neu.edu",
+    "maxEndDate": -1,
```
hankewyczz, line 7: Since `maxEndDate` is an optional field, we shouldn't initialize it unless we have real data for it. Applies for the whole file

scrapers/classes/parsersxe/tests/bannerv9Parser.test.ts
```diff
@@ -48,18 +48,21 @@ describe("getAllTermInfos", () => {
     expect(await bannerv9.getAllTermInfos("termslist")).toEqual([
       {
         host: "neu.edu",
+        maxEndDate: -1,
```
hankewyczz, line 51: Ditto

tests/database/dumpProcessor.test.seq.ts
```diff
@@ -31,11 +31,13 @@ const termInfos: TermInfo[] = [
     termId: "123456",
     subCollege: "NEU",
     text: "This is some text",
+    maxEndDate: -1,
```
hankewyczz, line 34: Ditto

types/types.ts
```diff
@@ -150,6 +150,7 @@ export interface TermInfo {
   termId: string;
   subCollege: string;
   text: string;
+  maxEndDate: number;
```
hankewyczz, line 153: This should reflect the actual type in Prisma
```suggestion
  maxEndDate?: number;
```

hankewyczz (changes requested):

graphql/resolvers/search.ts
```diff
@@ -1,7 +1,10 @@
 import { identity, pickBy } from "lodash";
 import searcher from "../../services/searcher";
 import { Course, Employee } from "../../types/types";
-import { AggResults } from "../../types/searchTypes";
+import { AggResults, SearchResult } from "../../types/searchTypes";
+import prisma from "../../services/prisma";
+import { TermInfo } from "@prisma/client";
+import { RssHandler } from "htmlparser2";
```
hankewyczz, line 7: Can be removed

PR #155, opened 2023-01-24
Add `POSTGRES_PASSWORD`

hankewyczz:
# Purpose

<br>

- https://hub.docker.com/_/postgres
- https://www.postgresql.org/docs/14/auth-trust.html

With our PSQL upgrade (#149), the default will throw an error if no Postgres password is provided. 

For local development, we don't use a password (this is not true for prod/staging). While we *could* override this setting with no issue, it's safer to default to a password. 

This only affects each user's local development environment, which is at very low-risk for access and/or misuse (all data is public anyways). 

# Tickets

- N/A

# Contributors

- @hankewyczz 

# Checklist

<br>
@sandboxnu/searchneu

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,6 @@ jobs:
         env:
           POSTGRES_MULTIPLE_DATABASES: searchneu_dev
           POSTGRES_USER: postgres
-          POSTGRES_PASSWORD: workflow_password
         ports:
           - 5432:5432
       elasticsearch:
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,7 @@ jobs:
         env:
           POSTGRES_MULTIPLE_DATABASES: searchneu_dev
           POSTGRES_USER: postgres
-          POSTGRES_PASSWORD:
+          POSTGRES_PASSWORD: workflow_password
         ports:
           - 5432:5432
       elasticsearch:
--- a/infrastructure/dev/docker-compose-server.yml
+++ b/infrastructure/dev/docker-compose-server.yml
@@ -2,7 +2,7 @@ version: "2"
 
 services:
   postgresql:
-    image: postgres:11.5
+    image: postgres:11.17-bullseye
     ports:
       - 5432:5432
     volumes:
@@ -11,9 +11,8 @@ services:
     environment:
       - POSTGRES_MULTIPLE_DATABASES=searchneu_dev,searchneu_test
       - POSTGRES_USER=postgres
-      - POSTGRES_PASSWORD=
   es:
-    image: docker.elastic.co/elasticsearch/elasticsearch-oss:7.1.0
+    image: docker.elastic.co/elasticsearch/elasticsearch-oss:7.10.2
     ports:
       - 9200:9200
     environment:
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -1,5 +1,5 @@
 generator client {
-  provider        = "prisma-client-js"
+  provider = "prisma-client-js"
 }
 
 datasource db {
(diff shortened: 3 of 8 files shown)


prisma/schema.prisma
```diff
@@ -1,5 +1,5 @@
 generator client {
-  provider        = "prisma-client-js"
```
hankewyczz, line 2: Automatic formatting changes

infrastructure/dev/docker-compose-server.yml
```diff
@@ -2,7 +2,7 @@ version: "2"
 
 services:
   postgresql:
-    image: postgres:11.5
+    image: postgres:11.17-bullseye
```
hankewyczz, line 5: Updated to match the other docker-compose file (which is more frequently used). This file is rarely (if ever) used, and only for local dev

.github/workflows/docker.yaml
```diff
@@ -15,7 +15,6 @@ jobs:
         env:
           POSTGRES_MULTIPLE_DATABASES: searchneu_dev
           POSTGRES_USER: postgres
-          POSTGRES_PASSWORD:
```
hankewyczz, line 18: By removing this, we allow the `POSTGRESS_PASSWORD` in the user's local environment to take precedence

template.env
```diff
@@ -6,4 +6,7 @@ TWILIO_AUTH_TOKEN=this_is_a_fake_variable
 TWILIO_VERIFY_SERVICE_ID=this_is_a_fake_variable
 
 CLIENT_ORIGIN=http://localhost:5000
-JWT_SECRET=
\ No newline at end of file
+JWT_SECRET=
+
+# This is only used for local development
+POSTGRES_PASSWORD=default_password
```
hankewyczz, line 12: To test these changes:
- delete your docker PSQL container
- `docker stop dev_postgresql_1 && docker rm dev_postgresql_1`
- add a password to your env file (can be anything)
- run `yarn dev:docker`
- run the scrape and updater again to populate the new database

Lucas-Dunker (approved):
Nothin to say

soulwa (changes requested):
current setup for passing environment variables doesn't fix the issue (at least, on my machine). 

there's a few ways to fix this, but the only one that I am aware of that keeps parity with the prisma environment is to add the `env_file` field to `docker-compose.yml` and have it reference the `.env` in the root of the project. only downside to this is polluting our container's environment (eg. the twilio variables would also get passed).

we could also pass it as an argument to `docker-compose up` when running `yarn dev:docker`, but this means you would need to always run the container through `yarn`. this is because the variables in `.env` are only local env variables when running a `yarn` commend.

i'll push something up which shows what the first solution would look like once I've verified that it works. would probably also be good to see that this works on at least another member's machine so that we know this is a more permanent solution.

hankewyczz:
You're right  — @Lucas-Dunker  could you please read this and re-review? I had an issue in my reproduction instructions (you also have to delete the Docker volume associated with the PSQL container).

> there's a few ways to fix this, but the only one that I am aware of that keeps parity with the prisma environment is to add the `env_file` field to `docker-compose.yml` and have it reference the `.env` in the root of the project. only downside to this is polluting our container's environment (eg. the twilio variables would also get passed).

I can confirm that this works. I'm not a massive fan of this behavior, but I don't see a significantly better approach.

Lucas-Dunker (approved):
Can confirm that this approach works on my machine as well, I'm able to make valid queries on the graphQL server like before; since the `docker-compose.yml` file now references the env file instead of a password, how/where does the container actually grab the password for authentication? Does it just automatically grab the password from the local environment?

hankewyczz:
> Can confirm that this approach works on my machine as well, I'm able to make valid queries on the graphQL server like before; since the `docker-compose.yml` file now references the env file instead of a password, how/where does the container actually grab the password for authentication? Does it just automatically grab the password from the local environment?

Yes and no — it inherits all environment arguments from the `.env` file, not necessarily the local environment (you can set environment variables outside of the `.env` file)

soulwa (approved):

soulwa:
new setup works for me, after nuking my docker setup and re-performing the backend setup process, so lgtm

sebwittr (approved):
Looks good to me

PR #162, opened 2023-02-27
Replace `request` with `got`

hankewyczz:
image


.github/workflows/test.yaml
```diff
@@ -9,6 +9,7 @@ jobs:
   tests:
     name: Tests
     runs-on: ubuntu-latest
+    timeout-minutes: 10
```
hankewyczz, line 12: This, and `detectOpenHandles`, were both added to handle the case that something doesn't properly close out. 

During this PR, I had issues where a dependency wouldn't kill itself, so Jest would stay open until Github forcibly killed it. 

This shouldn't be necessary anymore, but it's useful to keep just in case.

package.json
```diff
@@ -81,18 +81,19 @@
     "atob": "^2.0.3",
     "axios": "^0.25.0",
     "babel-core": "^7.0.0-bridge.0",
+    "cacheable-lookup": "^6.1.0",
```
hankewyczz, line 84: A replacement for `dnscache` that isn't deprecated, and actually works really well with `got`

package.json
```diff
@@ -81,18 +81,19 @@
...
     "core-js": "^3.0.1",
     "cors": "^2.8.5",
     "cross-env": "^7.0.3",
-    "dnscache": "^1.0.1",
     "domutils": "^2.8.0",
     "dotenv": "^16.0.0",
     "eslint-plugin-deprecation": "^1.2.1",
     "express": "^4.17.3",
     "fs-extra": "^10.0.0",
+    "got": "^11.8.2",
```
hankewyczz, line 96: Our new requests library!!!

package.json
```diff
@@ -111,6 +112,7 @@
     "request": "^2.81.0",
     "request-promise-native": "^1.0.9",
     "rollbar": "^2.19.4",
+    "tough-cookie": "^4.1.2",
```
hankewyczz, line 115: Used for storing cookies. Some URLs on the Banner site need specific cookies in order to work.

It goes something like this:
- you open the Banner site, select a specific term, and press "Submit"
- Banner saves a cookie essentially saying "this guy can search for classes in this term"

We want to save that cookie so that we can actually execute those searches. Our old requests library had this built in; the new one doesn't.

package.json
```diff
@@ -170,6 +172,7 @@
     "husky": "^4.3.8",
     "jest": "^27.0.0",
     "jest-environment-node": "^27.0.0",
+    "nock": "^13.3.0",
```
hankewyczz, line 175: This is used to mock requests while testing - this way, we don't have to rely on an Internet connection.

scrapers/classes/parsersxe/tests/classParser.test.js
```diff
@@ -17,10 +32,15 @@ beforeAll(() => {
   jest
     .spyOn(ClassParser, "getAttributes")
     .mockReturnValue(["innovation", "bizz"]);
+  jest.spyOn(ClassParser, "getFees").mockReturnValue({
```
hankewyczz, line 35: Before this PR, requests to this method were making real-life requests — something we want to avoid.

scrapers/classes/parsersxe/util.ts
```diff
@@ -94,25 +95,27 @@ async function getCookiesForSearch(termId: string): Promise<req.CookieJar> {
       form: {
         term: termId,
       },
-      cache: false,
+      cacheRequests: false,
+      cookieJar: cookieJar,
```
hankewyczz, line 99: we now have to manually handle the cookie jar - see my comment in `package.json` for more details

scrapers/main.ts
```diff
@@ -39,7 +40,13 @@ class Main {
       currentTermInfos: currentTermInfos,
     });
 
-    macros.log("Done scraping\n\n".green.underline);
+    const totalTime = Date.now() - start;
```
hankewyczz, line 43: This has nothing to (directly) do with the requests - I was just using this to measure the time difference between requests before and after my changes. 

I'm leaving this in because it's a feature that's nice to have.

scrapers/tests/request.test.ts
```diff
@@ -0,0 +1,28 @@
+/*
+ * This file is part of Search NEU and licensed under AGPL3.
```
hankewyczz, line 2: File was renamed from .js

The only major changes are the use of `httpbin.org` for our requests instead of using mocks. We do want to try an actual, live request at least once.

services/notifyer.ts
```diff
@@ -47,7 +47,8 @@ export async function sendNotifications(
     const courseNotifPromises: Promise<void>[] = notificationInfo.updatedCourses
       .map((course) => {
         const courseMessage = generateCourseMessage(course);
-        return courseHashToUsers[course.courseHash].map((user) => {
+        const users = courseHashToUsers[course.courseHash] ?? [];
```
hankewyczz, line 50: Handles the edge case where `courseHashToUsers` doesn't have any key matching `course.courseHash`. 

This isn't related to requests, just something I saw and fixed.

tests/general/utils/macros.test.ts
```diff
@@ -334,7 +334,7 @@ describe("console statements", () => {
       macros.error("test");
 
       // Called twice per macros.error call
-      expect(console.error).toHaveBeenCalledTimes(12);
+      expect(console.error).toHaveBeenCalledTimes(6);
```
hankewyczz, line 337: We no longer use `console.trace` in our macros calls, so the number of calls has been halved. The way `console.trace` is used in Node is now different, and it provides us with nothing useful.

scrapers/tests/__mocks__/request-promise-native.js
```diff
@@ -1,67 +0,0 @@
-/*
```
hankewyczz, line 1: This was used to mock the requests library for unit testing. 

We now use `nock` for that purpose, so this is no longer useful.

scrapers/request.ts
```diff
@@ -353,7 +318,8 @@ class Request {
     this.openRequests++;
 
     try {
-      return await request(output);
+      const { default: got } = await import("got");
```
hankewyczz, line 321: Add comment about ESM

pranavphadke1 (approved):

Lucas-Dunker (approved):
Looks great! For general knowledge transfer purposes, I'll shoot you a DM on slack so we can go through the PR together sometime if that works for you

scrapers/classes/parsersxe/bannerv9Parser.ts
```diff
@@ -19,7 +19,7 @@ import { ParsedCourseSR, ParsedTermSR } from "../../../types/scraperTypes";
 import { MultiProgressBars } from "multi-progress-bars";
 
 // Only used to query the term IDs, so we never want to use a cached version
-const request = new Request("bannerv9Parser", { cache: false });
+const request = new Request("bannerv9Parser", { cacheRequests: false });
```
Lucas-Dunker, line 22: Is `cacheRequests` just the term that `got` uses rather than `cache`?

scrapers/classes/parsersxe/termParser.ts
```diff
@@ -185,24 +186,25 @@ class TermParser {
...
-    const cookiejar = await util.getCookiesForSearch(termId);
+    const cookieJar = await util.getCookiesForSearch(termId);
     // second, get the total number of sections in this semester
     try {
       return (await this.concatPagination(async (offset, pageSize) => {
         const req = await request.get(
           "https://nubanner.neu.edu/StudentRegistrationSsb/ssb/searchResults/searchResults",
           {
-            qs: {
+            searchParams: {
```
Lucas-Dunker, line 196: I assumed that this change is just due to the way that `got` names its params

package.json
hankewyczz: `express` is a web framework, not really a requests library. 

Between requests libraries — [this comparison sums it up](https://github.com/sindresorhus/got#comparison). Obviously, that's a biased source (since it's from `got`), but it's not wrong. Essentially:
- `got` isn't a **bad** choice. It's in active development and very widely used. 
- It supports everything we need, and some things we might want to include in the future (eg. RFC-compliant caching)
- It's pretty simple to use

We could've gone with something else with no problem, but `got` just ended up being the choice.

scrapers/classes/parsersxe/bannerv9Parser.ts
hankewyczz: No — `cache` and `cacheRequests` are both our own options. However, `got` has its own `cache` option, and I didn't want people to get confused about this - is it `got`'s `cache` or our own?

scrapers/classes/parsersxe/subjectAbbreviationParser.ts
```diff
@@ -33,14 +33,12 @@ async function requestSubjects(termId: string): Promise<SubjectDescription[]> {
...
-  const response = await request.get(subjectUrl, {
-    json: true,
-  });
+  const response = await request.get(subjectUrl);
 
   if (response.statusCode !== 200) {
     macros.error(`Problem with request for subjects ${subjectUrl}`);
   }
-  return response.body;
+  return JSON.parse(response.body);
```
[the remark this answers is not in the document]
hankewyczz: Sure! Basically, our old library parsed the body to JSON if we specified `json: true`. This library _can_ do the same, but:

- Our requests library always returned the body of the request. We don't want to change this behavior (at least not in the PR) because it'd be a large breaking change
- In `got`, the `json()` method can only be called on a `Response` object. This object contains everything - we don't want to return that.

So the issue becomes - how can we convert to json without exposing the `Response` object and breaking compatibility?

This was my approach. In theory - we could add a `json` option, and then inside the request function, call `JSON.parse` on the response body if that option is true. That might be better - I'm open to changes. My initial thinking was - let's make this simple. A body is always a string, and after that, you do what you want with it.

scrapers/classes/parsersxe/termParser.ts
hankewyczz: I vividly remember writing a comment SOMEWHERE about this, but I can't find it. Oh well. That's exactly it. Read more [here](https://github.com/sindresorhus/got/blob/e24b89669931b36530219b9f49965d07da25a7e6/documentation/migration-guides/request.md#renamed-options)

scrapers/classes/parsersxe/tests/bannerv9Parser.test.ts
```diff
@@ -7,6 +7,29 @@ import prisma from "../../../../services/prisma";
...
+    },
+    {
+      code: "2",
+      description: "Summer 2022 Law Semester",
+    },
+  ])
+  .persist();
+
+afterAll(() => {
+  scope.persist(false);
```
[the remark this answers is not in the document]
hankewyczz: > When you setup an interceptor for a URL and that interceptor is used, it is removed from the interceptor list. This means that you can intercept 2 or more calls to the same URL and return different things on each of them. It also means that you must setup one interceptor for each request you are going to have, otherwise nock will throw an error because that URL was not present in the interceptor list. If you don’t want interceptors to be removed as they are used, you can use the [.persist()](https://github.com/nock/nock#persist) method.

That's why we set `persist()` at first - we want `nock` to use this interceptor for every matching request we have in the file.

However - we don't want this affecting other files! So at the end:
> If you want to stop persisting an individual persisted mock you can call persist(false):

scrapers/classes/parsersxe/tests/subjectAbbreviationParser.test.js
```diff
@@ -1,5 +1,24 @@
...
+    {
+      code: "AVM",
+      description: "Adv Manufacturing System - CPS",
+    },
+  ])
+  .persist();
+
+afterAll(() => {
+  scope.persist(false);
+});
```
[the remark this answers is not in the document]
hankewyczz: See above comment - does that answer it?

scrapers/classes/parsersxe/bannerv9Parser.ts
sebwittr: Is there any place where cacheRequests is being passed in as true?

scrapers/classes/parsersxe/subjectAbbreviationParser.ts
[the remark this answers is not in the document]
sebwittr: I think either option is fine. Personally I think it would be easier to have a json option.

sebwittr (approved):
Everything looks really good! The comments helped a lot with understanding this PR, nice job!

scrapers/classes/parsersxe/bannerv9Parser.ts
sebwittr: Disregard that, im being dumb^

scrapers/classes/parsersxe/subjectAbbreviationParser.ts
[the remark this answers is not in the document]
hankewyczz: Sure! Like I said — that's an open issue, could be discussed. Either way, this isn't a change that should be done in this PR, because that'll make this already too-complicated PR insanely overcomplicated

PR #164, opened 2023-03-12
Fix Prisma PSQL password access

hankewyczz:
Without these environment changes, Prisma won't work as intended. 

Steps:
1. Ensure that you have `POSTGRES_PASSWORD` in your `.env` file. This can be anything you want.
2. Log into Postgres.
- run `docker exec -it dev_postgresql_1 psql -U postgres`
- run `ALTER USER postgres WITH PASSWORD '<INSERT YOUR PASSWORD HERE>';`
3. Run `yarn db:migrate`
4. Run `yarn db:refresh`

Now, run `yarn scrape` to make sure everything's working as intended (this can be run with the local cache, the important thing is that it connects to PSQL)

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -15,7 +15,6 @@ jobs:
         env:
           POSTGRES_MULTIPLE_DATABASES: searchneu_dev
           POSTGRES_USER: postgres
-          POSTGRES_PASSWORD:
         ports:
           - 5432:5432
       elasticsearch:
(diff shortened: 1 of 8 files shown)


soulwa (approved):
works for me!

Lucas-Dunker (approved):
Works for me as well! In case other people are having trouble, running the first command didn't work for me at first since my docker container was named "dev-postgresql-1", not "dev_postgresql_1", so make sure your name matches up

pranavphadke1 (approved):

PR #169, opened 2023-03-29
Break out `.zip` file into individual files

hankewyczz:
# Purpose

Under the `data` directory, we currently have a .zip file containing a cache of the terms 202240, 202250, and 202260 (Northeastern Summer I, II, and full for 2022).

That .zip file is used in the end-to-end tests that run every time you commit to a PR. This is so we can test the cache — can we reliably fetch known data from the cache?

When updating the requests library, it'll be necessary to make some changes to the cache. I want to break this cache out from its zip file into individual files, so that you can actually see the diff. 

# Tickets

- https://trello.com/c/wKeHrjJH/314-replace-request-library

# Feature List

_Expand on the purpose - what does this code do, and how? This should be a list of the changes, more in-depth and technical_

- Break out zip files
- Scrape the latest data — these semesters are now archived, so this should be the last time we need to update the counts

# Reviewers

Secondary reviewers:

**Secondary**: @sebwittr


.github/workflows/test.yaml
```diff
@@ -86,9 +86,7 @@ jobs:
 
       - uses: bahmutov/npm-install@v1
 
-      - run: yarn add unzip
-
-      - run: unzip ./data/cache_2022_summer.zip
+      - run: mv ./data/cache_2022_summer ./cache
```
Lucas-Dunker, line 89: It seems we don't need to move the cache_2022_summer file to `./cache` in the original file, why is that?

hankewyczz: Originally, the top-level contents of the directory was just a single directory named "cache".

So, when you unzip the file, you're just left with that directory named "cache", which is unzipped in your current directory (ie. the directory from which you ran the command). 

Basically - this relies on the fact that the top-level dir in the zip is named "cache".

Lucas-Dunker (approved):

PR #172, opened 2023-04-02
Change bulkUpserter retry period and prevent erroring silently

sebwittr:
# Purpose

Our bulkIndex was failing silently, so now it will catch and log the errors. We also slightly increased the retry period in case it needed more time. 

# Tickets

# Contributors

@sebwittr 

Use the **"Assignees"** feature in Github

# Feature List

_Expand on the purpose - what does this code do, and how? This should be a list of the changes, more in-depth and technical_

- Increased retry period to 1000 to bulkIndex
- Added a try/catch to see if it failed, and logged it
-

# Notes (Optional)

_Is there anything reviewers should note? Do we need to add something later? Any followups? Was this ticket easier/harder than predicted?_

# Reviewers

**Primary**:

@hankewyczz  @Lucas-Dunker


utils/elastic.ts
```diff
@@ -257,13 +257,19 @@ export class Elastic {
...
           bulk.push(map[id]);
         }
-        // assumes that we are writing to the ES index name, not the ES alias (which doesn't have write privileges)
-        const res = await this.retryBulkQuery(indexName, bulk);
-
-        macros.log(
-          `indexed ${chunkNum * BULKSIZE + chunk.length} docs into ${indexName}`
-        );
-        return res;
+        try {
```
hankewyczz, line 260: We should toss this functionality into `retryBulkQuery`

hankewyczz (changes requested):

Lucas-Dunker (approved):
LGTM 🙏

utils/elastic.ts
sebwittr: retryBulkQuery already has some try catch functionality, and bulkIndexFromMap was the one that was silently erroring.

utils/elastic.ts
```diff
@@ -318,6 +319,11 @@ export class Elastic {
         return await client.bulk({ index: indexName, body: bulk });
       } catch (e) {
         macros.log(`Caught while bulk upserting: ${e.name} - ${e.message}`);
+
+        if (i == MAX_RETRY_ATTEMPTS - 1) {
```
hankewyczz, line 323:
```suggestion
        if (i === MAX_RETRY_ATTEMPTS - 1) {
```

hankewyczz (approved):

PR #173, opened 2023-04-05
Remove manual bulk upsert from dumpProcessor

hankewyczz:
# Purpose

We used to be manually hand-crafting SQL queries, which is easy to break and a security issue. All around - a messy piece of our data processing. This replaces that. 

# Tickets

- https://trello.com/c/mZ3B6bYS/350-replace-bulkupsert-mechanism-for-dumpprocessor

# Notes (Optional)

_Is there anything reviewers should note? Do we need to add something later? Any followups? Was this ticket easier/harder than predicted?_

An "upsert" is a combination of an "insert" and an "update" in SQL. 
- If the object already exists - update it
- If the object doesn't already exist - create it

When we process our data, we want to "upsert" our courses and sections. However, Prisma doesn't have a way to do this in bulk.

Currently, we are using a hand-crafted query. This can run upserts on multiple objects in a single query. However, this is fragile, isn't type-safe, and is generally frowned upon as bad practice. 

Instead, we can use `prisma.upsert` to upsert them one-by-one. This comes at a performance cost - this operation takes 60-80% longer (from my basic tests). However, we don't care!

- The scraper runs once a day, so as long as upserting takes less than 24 hours, this doesn't matter
- The updater runs every 5 minutes - same thing as above

There's a caveat in the case of the updater - the updater sends out SMS notifications for seats becoming free. Obviously, we want to send these out as fast as possible. As part of this PR, I changed the order of execution — notifications are now sent BEFORE the data is processed & inserted into the database and Elasticsearch.

# Reviewers

Secondary reviewers:

**Secondary**:
@pranavphadke1 @Lucas-Dunker


services/dumpProcessor.ts
```diff
@@ -4,29 +4,15 @@
...
-  Dump,
-  Section,
-  TransformFunction,
-} from "../types/types";
-
-type Maybe<T> = T | null | undefined;
+import { Dump, Section } from "../types/types";
 
 class DumpProcessor {
-  CHUNK_SIZE: number;
```
hankewyczz, line 24: We stopped using this a while ago, and just never removed it.

services/dumpProcessor.ts
```diff
@@ -348,15 +213,17 @@ class DumpProcessor {
...
+    const { desc, college, ...finalCourse } = correctedQuery;
 
     return finalCourse;
   }
 
   constituteSection(
     secInfo: Section,
     coveredTerms: Set<string>
-  ): Prisma.SectionCreateInput & { classHash: string } {
+  ): Prisma.SectionUncheckedCreateInput {
```
hankewyczz, line 226: It turns out Prisma has a type that includes the `classHash`, so we should use it instead of making up our own type.

services/updater.ts
```diff
@@ -113,6 +113,12 @@ class Updater {
       this.SECTION_MODEL
     );
 
+    await sendNotifications(
```
hankewyczz, line 116: Send notifications BEFORE the dump processor, so that users get the quickest possible notification.

tests/database/dbTestEnv.ts
```diff
@@ -17,7 +18,8 @@ class PrismaTestEnvironment extends NodeEnvironment {
     // Generate a unique schema identifier for this test context
     this.schema = `test_${uuid()}`;
     // Generate the pg connection string for the test schema
-    this.databaseUrl = `postgres://postgres@localhost:5432/searchneu_test?schema=${this.schema}`;
+    dotenv.config();
+    this.databaseUrl = `postgresql://postgres:${process.env.POSTGRES_PASSWORD}@localhost:5432/searchneu_test?schema=${this.schema}`;
```
hankewyczz, line 22: This should have been updated when I made the PSQL password changes. This only affects testing in dev environments — our tests on github were unaffected

tests/general/utils.test.ts
```diff
@@ -1,50 +0,0 @@
-import dumpProcessor from "../../services/dumpProcessor";
```
hankewyczz, line 1: I deleted all of the functions being tested in this file, so it's now useless to us.

services/updater.ts
```diff
@@ -70,7 +70,7 @@ class Updater {
   start(): void {
     // 5 min if prod, 30 sec if dev.
```
hankewyczz, line 71: change time in comment

services/dumpProcessor.ts
```diff
@@ -163,13 +74,26 @@ class DumpProcessor {
...
+            update: prismaSection,
+            where: {
+              id: prismaSection.id,
+            },
+          });
+        }
       );
+
+      // Execute in parallel
+      await Promise.all(upsertQueries);
```
tiingweii-shii, line 96: Abstraction 👾

tiingweii-shii (changes requested):

services/dumpProcessor.ts
hankewyczz: I tried today — you can't abstract this pattern without sacrificing Prisma's type saftey and using a few `any` types, which I'd rather avoid.

services/updater.ts
hankewyczz, line 71:
```suggestion
    // 5 min if prod, 1 min if dev.
```

services/updater.ts
sebwittr: Doesn't some of the data from the dump processor need to be used for the notifications?

sebwittr (approved):
Looks good for the most part, just had a question about doing notifications before the dumpProcessor

services/dumpProcessor.ts
```diff
@@ -137,18 +39,27 @@ class DumpProcessor {
...
-          courseTransforms,
-          courses.map((c) => this.constituteCourse(c, coveredTerms))
-        )
-      );
+    const processedCourses = termDump.classes.map((c) =>
+      this.constituteCourse(c, coveredTerms)
+    );
+    // Break the classes into groups of 2,000 each. Each group will be processed in parallel
+    // We can't process ALL classes in parallel because this may overwhelm the DB
+    const groupedCourses = _.chunk(processedCourses, 2000);
```
Lucas-Dunker, line 47: Do we still wanna use lodash here, or should we start [using the natural JS replacements](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore)?

services/dumpProcessor.ts
```diff
@@ -39,92 +25,8 @@ class DumpProcessor {
...
-      opt_prereqs_for: this.jsonTransform,
-      prereqs: this.jsonTransform,
-      prereqs_for: this.jsonTransform,
-      pretty_url: this.strTransform,
-      subject: this.strTransform,
-      term_id: this.strTransform,
-      url: this.strTransform,
-    };
-
-    const courseCols = [
```
Lucas-Dunker, line 68: What was this constant (and the others like it) used for, and why don't we need them anymore?

Lucas-Dunker: Additionally, what does `strTransform` and the other similarly named methods do?

Lucas-Dunker (changes requested):
Looks great to me aside from a few questions, god bless prisma upsert 🙏

services/dumpProcessor.ts
hankewyczz: I'm fine with leaving that in for now - [related reading](https://franciscov.medium.com/dont-let-the-boy-scouts-rule-hinder-pull-requests-8fdfbd4d5686) which sums it up nicely imo

services/dumpProcessor.ts
hankewyczz: This is used for preparing all of the entries for the upsert. 

Here's the original query: https://github.com/sandboxnu/course-catalog-api/blob/785b0cd429a9cb9f5d80ac6bc8a10ee6743da4ed/services/dumpProcessor.ts#L272-L294

Basically, the `transforms` methods are used to prepare the variables for the SQL query. Normally, Prisma does this for us, but here, we had to manually prepare everything, escape all the strings, convert non-string variables (dates, arrays) to their corresponding SQL string representation.

Lucas-Dunker (approved):
👍

PR #176, opened 2023-04-10
Add null checks to GraphQL resolvers

hankewyczz:
# Purpose

We sometimes get errors like `Cannot read property 'lastUpdateTime' of null`, which are unintuitive. This PR adds null checks to our GraphQL resolvers. 

If a user makes an API call for a course that doesn't exist, we should give them a descriptive error message.  

# Tickets

- N/A

# Reviewers

Primary reviewer:

- Pick one primary reviewer
  - The team member with the most relevant knowledge about the code area this PR touches
  - **NOT** an author of the PR
  - If the primary reviewer is the project lead, _select two primary reviewers_
    - Goal: facilitate knowledge transfer to other team members
  - Primary reviewers **are required to approve the PR** before it can be merged

**Primary**:

Use the **"Reviewers"** feature in Github

Secondary reviewers:

**Secondary**: @Lucas-Dunker

### DIFF
--- a/graphql/resolvers/class.ts
+++ b/graphql/resolvers/class.ts
@@ -10,6 +10,7 @@ import {
   Section as PrismaSection,
 } from "@prisma/client";
 import { Course, Section } from "../../types/types";
+import { GraphQLError } from "graphql";
 
 const serializer = new HydrateCourseSerializer();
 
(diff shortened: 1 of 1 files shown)


graphql/resolvers/class.ts
```diff
@@ -60,30 +73,63 @@ const getClassOccurrence = async (
...
     where: {
       uniqueCourseProps: { subject, classId, termId },
     },
     include: { sections: true },
   });
 
-  return serializeValues([res])[0];
+  if (result === null) {
+    throw new GraphQLError(
+      "We couldn't find a course matching the given term, subject, and class ID",
```
shashjar, line 85: Include subject, classId, and termId in GraphQL error?

shashjar: Nit but would be nice to standardize message structure with above instance

shashjar (approved):
Minor comment added, otherwise LGTM

PR #177, opened 2023-04-11
Pp/mark past terms inactive

pranavphadke1:
# Purpose

Right now we scrape past semesters while we only care about current and future semesters. This pr adds an "active" field to each termInfo (schema representing a term).

# Tickets

- https://trello.com/c/CQ0p1Ms4/391-mark-past-terms-as-inactive

# Contributors
@pranavphadke1 

# Feature List

- added an "active" field to termInfo prisma schema

# Reviewers

**Primary**:
@hankewyczz @sebwittr 

**Secondary**:
@Lucas-Dunker

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -121,6 +121,7 @@ model TermInfo {
   termId     String @unique(map: "term_ids.term_id_unique") @map("term_id")
   subCollege String @map("sub_college")
   text       String
+  active     Boolean
 
   @@map("term_ids")
 }
--- a/scrapers/classes/parsersxe/termListParser.ts
+++ b/scrapers/classes/parsersxe/termListParser.ts
@@ -12,7 +12,7 @@ class TermListParser {
     const activeTermInfos = termsFromBanner.filter(
       (term) => !term.description.includes("View Only")
     );
-    const activeTermIds = activeTermInfos.map(t => t.code);
+    const activeTermIds = activeTermInfos.map((t) => t.code);
     const minActiveTermId = Math.min(...activeTermIds);
     return termsFromBanner.map((term) => {
       const subCollege = this.determineSubCollegeName(term.description);
--- a/scrapers/classes/parsersxe/termListParser.ts
+++ b/scrapers/classes/parsersxe/termListParser.ts
@@ -9,13 +9,11 @@ class TermListParser {
   serializeTermsList(
     termsFromBanner: { code: string; description: string }[]
   ): TermInfo[] {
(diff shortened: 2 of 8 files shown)


prisma/schema.prisma
```diff
@@ -121,6 +121,7 @@ model TermInfo {
   termId     String @unique(map: "term_ids.term_id_unique") @map("term_id")
   subCollege String @map("sub_college")
   text       String
+  active     Boolean
```
hankewyczz, line 124:
```suggestion
  active     Boolean @default(true)
```

If you read through the errors in the migration page:
> Added the required column `active` to the `term_ids` table without a default value. This is not possible if the table is not empty.

Basically, the field you're adding needs to be non-null (which makes sense). However, there's no default value. So, when we apply these changes, it won't work, because there's existing data in the termInfos table!

Ping me when you start working on this, bc Prisma is tricky with undoing migrations. We can't just add a new migration, because that'd still try running the old one, which will still have the same issue.

scrapers/classes/parsersxe/termListParser.ts
```diff
@@ -9,6 +9,13 @@ class TermListParser {
   serializeTermsList(
     termsFromBanner: { code: string; description: string }[]
   ): TermInfo[] {
+    const activeTerms = termsFromBanner.filter(
+      (term) => !term.description.includes("View Only")
+    );
+    const minActiveCode = activeTerms.reduce((prev, cur) =>
+      prev.code < cur.code ? prev : cur
+    ).code;
+
```
hankewyczz, line 18:
```suggestion
    const activeTermInfos = termsFromBanner.filter(
      (term) => !term.description.includes("View Only")
    );
    const activeTermIds = activeTermInfos.map(t => t.code);
    const minActiveTermId = Math.min(...activeTermIds);
```
The code you have is fine, but I'd recommend a few small tweaks:
- The distinction between term ID (eg. `"202240"`) and termInfo (ie. an object containing the term ID, the term name, and other information) is annoying to keep track of
  - We want to explicitly define if something is a termInfo or a term ID
  - `code` isn't a term we use (even though Banner does) so we want to refer to it as a termID instead
- We want the code to be pretty simple & easy to understand. I'd argue that using `Math.min` makes it immediately apparent what's going on, while `reduce` requires that users actually read through the code to understand it

Again — all of this is fine & should work perfectly, but for clarity & making future onboarding easier, I'd prefer more explicit code

hankewyczz (commented):
Looking good! A few small tweaks, but nothing major

PR #178, opened 2023-04-13
Pp/mark past terms inactive 2

pranavphadke1:
# Purpose

Right now we scrape past semesters while we only care about current and future semesters. This pr adds an "active" field to each termInfo (schema representing a term).

# Tickets

- https://trello.com/c/CQ0p1Ms4/391-mark-past-terms-as-inactive

# Contributors
@pranavphadke1 

# Feature List

- added an "active" field to termInfo prisma schema

# Reviewers

**Primary**:
@hankewyczz @sebwittr 

**Secondary**:
@Lucas-Dunker


scrapers/main.ts
```diff
@@ -21,8 +21,11 @@ class Main {
     const allTermInfos = await bannerv9parser.getAllTermInfos(
       bannerv9CollegeUrls[0]
     );
+    // Filter out any past semesters
+    const activeTermInfos = allTermInfos.filter((t) => t.active);
+
```
pranavphadke1, line 26: Wasn't sure if I should do this initial filtering right now. This is the only place that actually filters out the past semesters. Thoughts? @hankewyczz

pranavphadke1:
idk how to pass the end-to-end tests

scrapers/main.ts
hankewyczz: No, we shouldn't do it here. I actually have some work I need to do on refactoring how `currentTermInfos` gets passed around, so I want to finish that before you make the changes here. In the meantime — can you add test support for this feature? Ping me if you need anything.

PR #187, opened 2023-05-08
Remove hacky Typescript conversions when working with Prisma

hankewyczz:
# Purpose

When working with Prisma, a common pattern we had littered around was `variable as unknown as Prisma.<type>`. 

Essentially, Typescript wouldn't let us cast to `Prisma.<type>`, so casting to `unknown` first allowed us to _force_ the conversion.

Needless to say, that's a bad idea. This removes that behavior by adding explicit conversion functions

# Tickets

- N/A

# Reviewers

Secondary reviewers: @soulwa

### DIFF
--- a/types/types.ts
+++ b/types/types.ts
@@ -3,10 +3,53 @@
  * See the license file in the root folder for details.
  */
 
-// A block of meetings, ex: "Tuesdays+Fridays, 9:50-11:30am"
 import { ParsedTermSR } from "./scraperTypes";
 import { Prisma } from "@prisma/client";
+import keys from "../utils/keys";
 
+/**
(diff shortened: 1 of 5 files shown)


types/types.ts
```diff
@@ -108,16 +108,16 @@ export interface Course {
 // A co or pre requisite object.
 export type Requisite = string | BooleanReq | CourseReq;
 
-export interface BooleanReq {
+export type BooleanReq = {
```
hankewyczz, line 111: This helps us convert types. 

Right now, something of type `CourseReq` is something which ONLY has the above keys. It's not a general dictionary/object in Typescript. 

When using `type` instead of `interface`, we can use the spread operator (`...`) to convert.

pranavphadke1 (approved):

PR #188, opened 2023-05-08
Break the updater into smaller functions for future refactoring

hankewyczz:
We'll need to do some refactoring in this file in the future. 

To make this easier — both to write and to understand — we'll want to break the updater into smaller functions first. That's all this PR does — breaks the functionality into smaller, bite-sized functions.

### DIFF
--- a/services/updater.ts
+++ b/services/updater.ts
@@ -91,12 +91,10 @@ class Updater {
     }
   }
 
-  // Update classes and sections users and notify users if seats have opened up
-  async update(): Promise<void> {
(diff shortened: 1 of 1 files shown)


Lucas-Dunker (changes requested):
Can you describe the purpose of this refactor in an little more depth? Correct me if im wrong, but this seems to be dividing/organizing all the scraping steps into function calls instead of just doing everything straight in the `updater.ts` file.

Lucas-Dunker (approved):

PR #189, opened 2023-05-08
Refactor how Section inputs are configured for Prisma

hankewyczz:
# Purpose

Before, we were using an _unchecked_ input for Prisma. Essentially, for each section input, we gave it a string (the class hash), and told Prisma "try to find this".

Now, we can use checked inputs - this uses actual input fields (classId, subject, termId) to determine if a class exists, and error elsewise. This is a cleaner solution which doesn't rely on string interpolation. 

Additionally, this PR moves the class hash checking. Essentially, this is the functionality that checks "does this section actually have a corresponding hash?". This sets us up for another PR. The gist of it is — we want to do this check in the `updater`, so that we can do some follow-up work here, instead of in the `dumpProcessor`.

# Tickets

- N/A

### DIFF
--- a/services/dumpProcessor.ts
+++ b/services/dumpProcessor.ts
@@ -36,6 +36,7 @@ class DumpProcessor {
       this.convertSectionToDatabaseFormat(section)
     );
     await this.saveSectionsToDatabase(processedSections);
+    await this.updateSectionsLastUpdateTime(termDump.sections);
 
     await this.saveSubjectsToDatabase(termDump.subjects);
 
@@ -132,7 +133,6 @@ class DumpProcessor {
     }
 
     macros.log("Finished with sections");
-    await this.updateSectionLastUpdateTime(sections);
   }
 
   /**
@@ -140,9 +140,7 @@ class DumpProcessor {
    * We use this to track sections that haven't been updated in a while,
    * which means that they're no longer on Banner & should be removed from our database.
    */
-  async updateSectionLastUpdateTime(
-    sections: Prisma.SectionCreateInput[]
-  ): Promise<void> {
+  async updateSectionsLastUpdateTime(sections: Section[]): Promise<void> {
     await prisma.course.updateMany({
       where: { id: { in: sections.map((s) => keys.getClassHash(s)) } },
       data: { lastUpdateTime: new Date() },
--- a/services/dumpProcessor.ts
+++ b/services/dumpProcessor.ts
@@ -107,7 +107,7 @@ class DumpProcessor {
    * Performs a SQL upsert - insert if the section doesn't exist, update if it does.
    */
   async saveSectionsToDatabase(
-    sections: Prisma.SectionUncheckedCreateInput[]
+    sections: Prisma.SectionCreateInput[]
   ): Promise<void> {
     const updateTime = new Date();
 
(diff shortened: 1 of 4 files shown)


hankewyczz:
Depends on #188

services/dumpProcessor.ts
```diff
@@ -32,17 +32,11 @@ class DumpProcessor {
...
-        (elem) => elem.id
-      )
+    const processedSections = termDump.sections.map((section) =>
+      this.convertSectionToDatabaseFormat(section)
     );
-    const processedSections = termDump.sections
-      .map((section) => this.convertSectionToDatabaseFormat(section))
-      .filter((s) => courseIds.has(s.classHash));
     await this.saveSectionsToDatabase(processedSections);
+    await this.updateSectionsLastUpdateTime(termDump.sections);
```
hankewyczz, line 39: I pulled this out from `saveSectionsToDatabase` because it makes more sense at the top-level (in hindsight). 

This operates over the raw sections (ie. not compatible with Prisma), not the processedSections

services/dumpProcessor.ts
```diff
@@ -139,19 +133,16 @@ class DumpProcessor {
...
 
   /**
    * Update the lastUpdateTime attribute on all given sections.
    * We use this to track sections that haven't been updated in a while,
    * which means that they're no longer on Banner & should be removed from our database.
    */
-  async updateSectionLastUpdateTime(
-    sections: Prisma.SectionUncheckedCreateInput[]
-  ): Promise<void> {
+  async updateSectionsLastUpdateTime(sections: Section[]): Promise<void> {
```
hankewyczz, line 143: Renamed to be plural

services/dumpProcessor.ts
```diff
@@ -139,19 +133,16 @@ class DumpProcessor {
...
     await prisma.course.updateMany({
-      where: { id: { in: sections.map((s) => s.classHash) } },
+      where: { id: { in: sections.map((s) => keys.getClassHash(s)) } },
```
Lucas-Dunker, line 145: What's the difference between this and `classHash`?

services/dumpProcessor.ts
```diff
@@ -287,19 +278,28 @@ class DumpProcessor {
...
+        connect: {
+          uniqueCourseProps: {
+            classId: secInfo.classId,
+            termId: secInfo.termId,
+            subject: secInfo.subject,
+          },
+        },
+      } as Prisma.CourseCreateNestedOneWithoutSectionsInput,
     };
     return _.omit({ ...secInfo, ...additionalProps }, [
```
Lucas-Dunker, line 297: What do underscores mean again for semantics?

services/dumpProcessor.ts
hankewyczz: The type of `sections` has been changed — essentially, we no longer have `classHash`, so we need to compute it. The values are identical — the only difference is that the new `sections` type doesn't already have `classHash` as a property

services/dumpProcessor.ts
hankewyczz: Conventionally, a variable prefixed with an underscore won't be used (eg. `_`, `_var`). 

HOWEVER, we also import lodash as an underscore. Not sure why. In this usage, the underscore is the lodash library

pranavphadke1 (approved):

Lucas-Dunker (approved):

PR #194, opened 2023-05-22
Upgrade AWS instances

hankewyczz:
# Purpose

Back in Nov-Dec 2022, we had to cut back on our AWS instances due to cost pressures. 

However, we shouldn't be running at these lower levels for a long time - we have way too many users for what we have to work properly. This is compounded by Graduate releasing their product to the public (congrats!!) so we need a few upgrades.

### DIFF
--- a/infrastructure/terraform/modules/course-catalog-api/elasticsearch.tf
+++ b/infrastructure/terraform/modules/course-catalog-api/elasticsearch.tf
@@ -12,7 +12,7 @@ module "elasticsearch" {
   availability_zone_count = 1
 
   elasticsearch_version   = "7.10"
-  instance_type           = "t2.small.elasticsearch"
+  instance_type           = "t3.small.elasticsearch"
   instance_count          = 1
   ebs_volume_size         = 20
   encrypt_at_rest_enabled = "false"
--- a/infrastructure/terraform/modules/course-catalog-api/rds.tf
+++ b/infrastructure/terraform/modules/course-catalog-api/rds.tf
@@ -4,7 +4,7 @@ resource "aws_db_instance" "default" {
   allocated_storage      = 20
   engine                 = "postgres"
   engine_version         = "11.17"
-  instance_class         = "db.t3.micro"
+  instance_class         = "db.t3.small"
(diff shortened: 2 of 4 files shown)


infrastructure/terraform/modules/course-catalog-api/elasticsearch.tf
```diff
@@ -12,7 +12,7 @@ module "elasticsearch" {
   availability_zone_count = 1
 
   elasticsearch_version   = "7.10"
-  instance_type           = "t2.small.elasticsearch"
+  instance_type           = "t3.small.elasticsearch"
```
hankewyczz, line 15: 2x the CPU, same cost (newer gen.)

infrastructure/terraform/modules/course-catalog-api/rds.tf
```diff
@@ -4,7 +4,7 @@ resource "aws_db_instance" "default" {
   allocated_storage      = 20
   engine                 = "postgres"
   engine_version         = "11.17"
-  instance_class         = "db.t3.micro"
+  instance_class         = "db.t3.small"
```
hankewyczz, line 7: 2x the mem - this has been a major issue for us. **This might require a follow-up boost to CPU as well!!!**. 

Cost increase: ~15/mo per instance, ~30/mo

infrastructure/terraform/modules/course-catalog-api/variables.tf
```diff
@@ -80,7 +80,7 @@ variable "webapp_cpu" {
 
 variable "webapp_memory" {
   description = "memory the webapp should get"
-  default     = 1024
+  default     = 3072
```
hankewyczz, line 83: Triple the memory for our API and updaters, both of which have struggled to keep up w/ increased usage of Search. 

Cost increase: ~28/mo (total)

sebwittr:
Upgrading everything makes sense. How long until we run out of funds with these changes though?

sebwittr (approved):

Lucas-Dunker (approved):
I LOVE TERRAFORM!!!!

kinda freaking out for this fall when funding becomes an issue again lol but rn we chillin

PR #195, opened 2023-05-24
By default, the scraper should only scrape new terms

hankewyczz:
# Purpose

Our scrapers used to be divided:
- Scraping scrapes class data
- Updating scrapes section data

Now, with #190, that line has been blurred. Once a term is scraped initially, the updater will automatically scrape any new classes _for that term_.

Therefore, there's no need to scrape a term more than once anymore (unless the class titles, descriptions, etc. change).

So, we should default our scrapers to only scrape __NEW__ terms. Of course, this can be overridden by several env. variables.

# Reviewers

Primary reviewer:

- Pick one primary reviewer
  - The team member with the most relevant knowledge about the code area this PR touches
  - **NOT** an author of the PR
  - If the primary reviewer is the project lead, _select two primary reviewers_
    - Goal: facilitate knowledge transfer to other team members
  - Primary reviewers **are required to approve the PR** before it can be merged

**Primary**:

Use the **"Reviewers"** feature in Github

Secondary reviewers:

**Secondary**:

### DIFF
--- a/scrapers/classes/tests/classScraper.test.ts
+++ b/scrapers/classes/tests/classScraper.test.ts
@@ -1,36 +1,62 @@
-import scraper, { NUMBER_OF_TERMS_TO_UPDATE } from "../../main";
+import scraper from "../../main";
+import prisma from "../../../services/prisma";
 
 describe("getTermsIds", () => {
   beforeEach(() => {
-    process.env.TERMS_TO_SCRAPE = "";
+    delete process.env.TERMS_TO_SCRAPE;
+    delete process.env.NUMBER_OF_TERMS;
   });
 
-  it("returns the termsStr if and only if they're in the terms list", () => {
+  it("returns the termsStr if and only if they're in the terms list", async () => {
     process.env.TERMS_TO_SCRAPE = "202210,202230,202250";
-    expect(scraper.getTermIdsToScrape([])).toEqual([]);
-    expect(scraper.getTermIdsToScrape(["202210"])).toEqual(["202210"]);
+    expect(await scraper.getTermIdsToScrape([])).toEqual([]);
+    expect(await scraper.getTermIdsToScrape(["202210"])).toEqual(["202210"]);
     expect(
-      scraper.getTermIdsToScrape(["202210", "202230", "202250", "1234"])
+      await scraper.getTermIdsToScrape(["202210", "202230", "202250", "1234"])
     ).toEqual(["202210", "202230", "202250"]);
   });
 
-  it("without a termStr, it takes NUMBER_OF_TERMS_TO_PARSE terms", () => {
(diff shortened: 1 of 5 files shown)


sebwittr (approved):
Changes look good to me!

scrapers/main.ts
```diff
@@ -14,47 +14,45 @@ import "colors";
...
+      return termIds.filter((t) => !termIdsWithData.has(t));
+    }
   }
 
   async main(): Promise<void> {
     const start = Date.now();
     // Get the TermInfo information from Banner
     const allTermInfos = await bannerv9parser.getAllTermInfos();
-    const termsToScrape = this.getTermIdsToScrape(
+    const termsToScrape = await this.getTermIdsToScrape(
```
Lucas-Dunker, line 55: Slightly unrelated, but is there ever even a time where we should use `await` for async calls like this? Was `await` not being in any of these just an optimization oversight?

scrapers/classes/tests/classScraper.test.ts
```diff
@@ -1,36 +1,62 @@
-import scraper, { NUMBER_OF_TERMS_TO_UPDATE } from "../../main";
+import scraper from "../../main";
+import prisma from "../../../services/prisma";
 
 describe("getTermsIds", () => {
   beforeEach(() => {
-    process.env.TERMS_TO_SCRAPE = "";
+    delete process.env.TERMS_TO_SCRAPE;
```
Lucas-Dunker, line 6: Why do we delete these environment variables for the tests?

scrapers/main.ts
```diff
@@ -14,47 +14,45 @@ import "colors";
...
       macros.log("Scraping using user-provided TERMS_TO_SCRAPE");
       return terms;
-    }
-
-    const rawNumTerms = Number.parseInt(process.env.NUMBER_OF_TERMS);
-    const numTerms = isNaN(rawNumTerms)
-      ? NUMBER_OF_TERMS_TO_UPDATE
-      : rawNumTerms;
+    } else if (!isNaN(numOfTermsStr)) {
+      return termIds.slice(0, numOfTermsStr);
```
Lucas-Dunker, line 40: In what case would this code be executed?

Lucas-Dunker (commented):
So, in summary, does this refactor make the default case of scraping that only new terms are scraped, but if environment variables are found, it scrapes for those terms instead?

scrapers/main.ts
hankewyczz: This wasn't an `async` call before - it is now

scrapers/main.ts
hankewyczz: If `numOfTermsStr` exists and is a number - ie. if someone set the environment variable `NUMBER_OF_TERMS` to a number.

Lucas-Dunker (approved):

PR #213, opened 2024-01-23
Upgrade core dependencies + dev node variable

Lucas-Dunker:
# Purpose

While I updated the frontend dependencies at the start of Fall 2023, I put off doing the same for the backend due to issues I had with updating Elasticsearch; it turns out that it was way easier to update than expected, and I just had to tweak my yarn installation a little bit. Now there's no longer a node version mismatch between both of our search repos, and while a few of our dependencies are still end-of-life, all of our core ones should be supported at least for the next few years.

# Tickets

[This one](https://trello.com/c/oii3iD9C/411-outdated-dependencies-on-cca)

# Contributors

@Lucas-Dunker 

# Feature List

- Update our node, babel, graphQL, and Elasticsearch dependencies, as well as their corresponding types
- Add a dev variable to `template.env` that I forgot about

# Notes 

**Changes like this are very prone to break things!!!** Ideally at least half of the team should verify that these changes work before deploying to staging and prod:

- Use the new node version in `.nvmrc` with `fnm use`
- Run `yarn` to get the new dependencies; if there's any errors, try deleting `node_modules` and rerunning the command
-  Run all tests with `yarn test`
- Try out a scrape using `yarn scrape`, ideally both a cached one and a live one
- Run a graphQL instance locally with `yarn dev`, and execute a few example queries to make sure everything is working as expected

# Reviewers

**Primary**:
@pranavphadke1 @sebwittr 

**Secondary**:
@soulwa @ananyaspatil @Anzhuo-W


pranavphadke1 (approved):

mehallhm:
PR superseded by #231

PR #216, opened 2024-04-13
Have better section filtering

sebwittr:
# Purpose

We have a bug where if you filter by "honors" and "campus", you won't guarantee that there will be one section with the given honors and campus. This will guarantee each class returned has one section matching both filters, and it filters out the rest of the sections as well. 

# Tickets

https://trello.com/c/v6dY3aCt/416-section-filters-should-only-return-classes-that-have-sections-that-match-both-filters

# Contributors

@ananyaspatil @sebwittr 

# Feature List

- Changed the structure of our ES mapping to make sections a nested property. This allows to us to have nested queries and maps the sections directly to the classes withing Elasticsearch. This is necessary for doing multiple section filters.
- Changed the current query to use a nested query for any section filters.
- Added some code to filter out any sections from the results that don't match all the filters.

# Notes (Optional)

The aggregations are still inaccurate (if you turn on the honors filter, and then click the campus filter you'll see that there are a lot of campus options that lead to no results after clicking it). We'll need to make a separate ticket for this!

# Reviewers

Primary reviewer: @pranavphadke1 @soulwa 

Secondary reviewers:
@ananyaspatil @Anzhuo-W 

Original results:
image
With the new filtering:
image

### DIFF
--- a/scrapers/classes/classMapping.json
+++ b/scrapers/classes/classMapping.json
@@ -41,7 +41,8 @@
         }
       },
       "sections": {
-        "type": "object",
+        "type": "nested",
+        "include_in_parent": "true",
         "properties": {
           "meetings": {
             "type": "object"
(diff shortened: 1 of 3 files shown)


sebwittr:
I'll fix the tests later.. oops

PR #218, opened 2024-08-15
Ap/unsubscribe after 3 notifs

ananyaspatil:
# Purpose
To reduce twilio costs, a user will be automatically unsubscribed from a course/section after they receive 3 text notifications for that course/section. 

# Tickets
https://trello.com/c/Aknz6Jdf 

# Contributors
@ananyaspatil @sebwittr @pranavphadke1 @elvincheng3 

# Feature List
- Add notif_count attribute to prisma database schema to track of # of notifs sent out per course or section subscription
- Send user a text notification if notif_count < 3 for a given course / section
- Delete the course/section subscription after 3 notifs limit reached
- Create 'unit' testing subdirectory, move unit tests from general to unit

# Reviewers

**Primary**:
@sebwittr @pranavphadke1 @Anzhuo-W @Lucas-Dunker 

Secondary reviewers:
@cherman23 @nickpfeiffer05 @mehallhm


.github/workflows/test.yaml
```diff
@@ -37,6 +37,10 @@ jobs:
           node-version: "16"
       - uses: bahmutov/npm-install@v1
 
+      - run: yarn db:migrate
```
elvincheng3, line 40: this yaml change wasn't intended to be a fix, just a proof of concept. i think it would make sense if we broke down the test directories into unit and integration subdirectories, and had separate testing CI for unit vs integration tests. it's best if we have clear separation between unit and integration, just in case we have tests that were intended to be unit tests but ended up relying on some dependency. would also be best if the unit tests did not have the db, elastic, or graph spun up.

elvincheng3 (commented):
made a suggestion, could be for a future ticket but i think it would be best addressed now so it doesn't get overlooked down the line.

services/updater.ts
```diff
@@ -527,6 +530,7 @@ class Updater {
     );
   }
 }
+export { Updater };
```
elvincheng3, line 533: why is this export needed?

tests/general/notifyer.test.ts
```diff
@@ -257,5 +460,497 @@ describe("Notifyer", () => {
         expectedSectionMessage
       );
     });
+
+    it("does not send any notifications for each course and section when each subscribed section and class has notifCount>=3", async () => {
+      console.log("TEST 6");
```
elvincheng3, line 465: let's clean up the console logs?

services/updater.ts
ananyaspatil: I was using Updater in the Notifyer tests, have edited the tests and removed this import.

sebwittr (approved):

PR #219, opened 2024-09-24
Upgrade Node and core dependencies

mehallhm:
# Purpose
This is a living PR to upgrade all major dependencies for the catalog api. In a nutshell, nearly every dependency for the project is either deprecated, end-of-life, or unmaintained; the goal is to bring all the dependencies to a more up-to-date version as well as make it easier to avoid this situation in the first place. This PR builds off the work of #213.

# Contributors
@mehallhm 

# Notes 
The primary goal will be to upgrade Node versions. Node 16 (the currently used version) was eol a year ago. The issue is that the version of elasticsearch currently used cannot be run on newer versions of Node. After that, upgrading all the other packages to newer versions should be simple, nothing else has such bad cross-dependence.

Secondary goals include minimizing dependencies through vendoring and more modern tooling to avoid this problem in the future as well as build system / tooling changes. A newer Node version allows native support of modern Java/Typescript syntax and language features without polyfills, potentially saving build complexity and allowing moves towards newer standards, such as ES Modules.

# Reviewers
This draft is meant to solicit feedback and create a conversation around the best way to move forward. As such, it is not ready to be reviewed in its entirety yet.


elvincheng3:
Would it be better to create an epic for updating dependencies, and then break each task into stories? I think it might be less risky to isolate each update instead of adding them into one large PR.

mehallhm:
It would probably be smart to do it as atomically as possible. However, this branch doesn't even run yet. I do agree that extensive testing is necessary before merging this mess - ideally we would do a side by side deployment and incrementally migrate over. Once the code can actually run, we could then chop it into smaller PRs and merge those progressively as well (kinda like a stacking workflow). For now, this is still a draft to actually get something working.

Issue #233, opened 2024-10-27
Campus Section Filter


Anzhuo-W:
Problem: Feedback from advisor: "When selecting campus under the search function, is there any way that only courses with faculty from that location can be listed? When I searched by Silicon Valley, it still showed other campuses sections (e.g. Seattle)." (ie if we are filtering classes by campus we should probably only show the sections for that campus)

Solution: If campus filters are applied, only display sections for that campus instead of all sections for a course where at least one section is on that campus.

Est. Time for Completion: 1 week

PR #234, opened 2024-10-27
Add section filters

Anzhuo-W:
# Purpose

In returned results, only show specific sections that match all filters of the classes with at least one section that matches filters.

Before:
Screen Shot 2024-10-27 at 2 59 24 PM

After:
Screen Shot 2024-10-27 at 3 00 23 PM

# Tickets

Addresses #233 

# Contributors

@Anzhuo-W 

# Feature List

When campus filters are applied, sections are only displayed for that campus instead of all sections for a course where at least one section is on that campus. The sections are filtered out from the results after they are returned by query; additionally applies to honors sections, class type, and subject.

# Reviewers

Primary reviewer:
@ananyaspatil 

Secondary reviewers:
@nickpfeiffer05 @cherman23 @mehallhm

### DIFF
--- a/services/searcher.ts
+++ b/services/searcher.ts
@@ -10,7 +10,6 @@ import {
 import prisma from "../services/prisma";
 import elastic, { Elastic } from "../utils/elastic";
 import HydrateSerializer from "../serializers/hydrateSerializer";
-import HydrateCourseSerializer from "../serializers/hydrateCourseSerializer";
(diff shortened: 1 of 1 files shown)


mehallhm (approved):
I wish there wasn't a way to iterate over everything multiple times, but I have no ideas rn.  Depending on merge order, this might lightly conflict with Nick's #235

PR #220, opened 2024-09-29
Refactor serializers

nickpfeiffer05:
# Purpose

Lodash is unnecessary and can be replaced by native features. Serializer methods are better suited as static methods. 

# Tickets

https://github.com/orgs/sandboxnu/projects/20?pane=issue&itemId=80651976

# Contributors

@Anzhuo-W , @nickpfeiffer05 

# Feature List

1. Removed reliance on Lodash in all serializer files
    - The previous serializers used many Lodash functions (.pick, .groupBy, etc.), that can now be replaced easily by modern JS features
    - PR has replaced all old Lodash functions with native JS
2. Conversion to static methods
    - Converted all serializer methods to static
    - Adjusted the use of serializers from instanced-calling to static-calling in `searcher.ts`

# Reviewers

**Primary**:

@ananyaspatil 

**Secondary**:

@mehallhm , @cherman23

PR #235, opened 2024-10-27
Test serializers

nickpfeiffer05:
# Purpose

Wrote unit and integration tests for the recently refactored serializers. We previously had no serializer tests, and these tests will ensure that any future changes to the serializers will not produce unintended results. 

# Contributors

@nickpfeiffer05 

# Feature List

- Added testing files `integration.test.ts` and `unit.test.ts` in the `serializers` testing folder. Produced sample course data for testing in the `fixtures` folder
- Wrote unit tests for the bulkSerialize method of all serializers
- Wrote integration tests for graphQL -> bulkSerialize method of all serializers. This ensures that all course data retrieved from database data is serialized correctly.
   
# Reviewers

Primary reviewer:

**Primary**:

@ananyaspatil 

Secondary reviewers:

**Secondary**:

@Anzhuo-W @mehallhm @cherman23


mehallhm (changes requested):
Instead of `unit.test.ts` and `integration.test.ts` could you please instead put each test type in the appropriate directory and renamed to something more descriptive to their role for the serializers. I understand that the fixtures need to be shared, so those can just live in the `tests` dir. That way our integration and unit tests don't get jumbled together

mehallhm:
Does this PR succeed #220? If so, can you please clean it up to prepare for the changes here.

mehallhm (approved):
Given what we discussed, we should move forward with the way it is now and refactor all of the tests at a later date

PR #237, opened 2024-10-31
Updater healthcheck

cherman23:
# Purpose

Currently there is no way to know if updater runs properly. With healthchecks a ping will be sent to the searchneu-support channel.

# Tickets

https://github.com/orgs/sandboxnu/projects/20?pane=issue&itemId=82334730

# Contributors

@cherman23 

# Feature List

- The updater is currently set to run every 5 minutes, a ping is sent out to healthchecks.io when the the updater runs properly. If the updater doesn't run the service will wait 10 minutes prior to sending a message to the searchneu-support channel in the sandbox slack.
- The service sends out a get request to link provided in the healthchecks.io account. 
- Any modifications made to how we get a notification if updater doesn't process must be done through healthchecks.io portal.

# Notes 

I @cherman23 have the login and am the owner for the healthchecks.io account and @ananyaspatil is a manager of the account. Contact either for passing off login information and/or changing ownership.

# Reviewers:

Primary reviewer:

@ananyaspatil 

Secondary reviewers:

@mehallhm 
@nickpfeiffer05 
@olivial215 
@serena-boyu

### DIFF
--- a/services/healthchecks.ts
+++ b/services/healthchecks.ts
@@ -1,4 +1,5 @@
 import axios from "axios";
+import macros from "../utils/macros";
 
 export async function sendHealthCheck(): Promise<void> {
   const pingURL = "	https://hc-ping.com/7d9197be-db67-4cc5-be00-1a3df56d0918";
@@ -9,6 +10,6 @@ export async function sendHealthCheck(): Promise<void> {
   try {
     await axios.get(pingURL);
   } catch (error) {
-    console.log("Health checks service is not responding properly " + error);
+    macros.warn("Health checks service is not responding properly: " + error);
   }
 }
--- a/services/healthchecks.ts
+++ b/services/healthchecks.ts
@@ -1,0 +1,14 @@
+import axios from "axios";
+
+export async function sendHealthCheck(): Promise<void> {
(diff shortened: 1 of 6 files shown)


services/healthchecks.ts
```diff
@@ -0,0 +1,15 @@
+import axios from "axios";
+import macros from "../utils/macros";
+
+export async function sendHealthCheck(): Promise<void> {
+  const pingURL = "	https://hc-ping.com/7d9197be-db67-4cc5-be00-1a3df56d0918";
```
mehallhm, line 5: Can we move the URL to an env variable? That way we avoid hard coding the URL here, making it easier to change and manage in the future while centralizing configuration to the env variables

services/updater.ts
```diff
@@ -362,6 +363,8 @@ class Updater {
     await this.sendUserNotifications(sections);
     // Save the data in our database
     await this.saveDataToDatabase(sections);
+    // Send out healthcheck ping
+    macros.PROD ? await sendHealthCheck() : null;
```
mehallhm, line 367: Can we rewrite this as a simple if statement? It increases readability and the second branch of the ternary is unused anyway

mehallhm (approved):

PR #241, opened 2024-11-13
Yarn berry

### LINKED ISSUE #225: Upgrade to Yarn Berry
Yarn classic (v1) entered maintenance mode in [January of 2020](https://classic.yarnpkg.com/lang/en/docs/install/#mac-stable). The recommendation is to upgrade to [Yarn Berry (v4) ](https://yarnpkg.com/getting-started). This process should not be terribly difficult from the repo side, but would require developer tooling updates. 

As Yarn Berry operates through [Corepack](https://github.com/nodejs/corepack), there would be friction in this transition and Corepack is not certain to remain bundled with Node moving forward. There is not a major reason to move off of Yarn to alternative package managers, though that option does exist.

### LINKED ISSUE #232: Upgrade all GitHub actions
As noted [here](https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/), many of the actions used in this repo use Node version 12 or Node version 16. These will be forced to run on Node 20, which they were not meant for and will eventually break once Node removes a feature they depend on. Upgrading to the latest version should mitigate these issues

mehallhm:
# Purpose
Migrates off of the end-of-life Yarn Classic to the actively maintained Yarn Berry version. In general, it is a drop in replacement, using the Node experimental [Corepack](https://nodejs.org/api/corepack.html) as the preferred installation / management tool (the irony of managing a package manager).

# Tickets
Closes #225, Closes #232

# Feature List
- Switches from Yarn Classic to Yarn Berry in `node_modules` compatibility mode.
- Updates the Dockerfile to enable Corepack and use the same version of Yarn as dev, ensuring consistency.
- Prepped Dockerfile for a better build system in the future. Yarn Berry forces us to stop abusing Yarn like we were before so some of this needed to happen now anyway. This will be something that will show its value soon...
- Updates the GH actions to use Corepack to install deps (and generally get them to newer versions).

# Notes
I will open a PR on [searchneu](https://github.com/sandboxnu/searchneu) soon to ensure that there is a version locked in because Corepack likes to add itself otherwise. 

# Reviewers
**Primary**:
@ananyaspatil 

**Secondary**:
@Anzhuo-W @nickpfeiffer05 @cherman23


mehallhm:
I'm really not liking these `request` unit tests... They are super finicky both in CI and locally. I can barely `curl` these URLs! Something we should improve....

mehallhm:
This PR has been merged into #244 to fix a Dependency issue preventing the CI from running

PR #280, opened 2025-03-18
update prisma schema for improved notification logging

cherman23:
# Purpose

Update prisma schema for improved logging considering notifications

# Contributors

@cherman23 

# Feature List

- Add table for tracking sent notifications
- Add table for when notifications are updated
- update followed sections 

# Reviewers

Primary reviewer:

@ananyaspatil 
@mehallhm 

Secondary reviewers:

@ItsEricSun 
@WesleyTran0 
@nickpfeiffer05


cherman23:
originally branched off of old master

prisma/schema.prisma
```diff
@@ -100,13 +100,16 @@ model FollowedCourse {
...
-  notifCount  Int     @default(0) @map("notif_count")
-  section     Section @relation(fields: [sectionHash], references: [id], onDelete: Cascade)
-  user        User    @relation(fields: [userId], references: [id], onDelete: Cascade)
-
-  @@id([userId, sectionHash])
+  sectionHash String    @map("section_hash")
+  userId      Int       @map("user_id")
+  notifCount  Int       @default(0) @map("notif_count")
+  created_at  DateTime  @default(now())
+  updated_at  DateTime  @default(now())
```
mehallhm, line 107: There is a prisma `updatedAt` which automagically updates the field on every write

prisma/schema.prisma
```diff
@@ -91,7 +91,7 @@ model Subject {
 model FollowedCourse {
   courseHash String @map("course_hash")
```
mehallhm, line 92: I know we only mentioned it in the meeting regarding sections, but all the new fields on the FollowedSection are also applicable to the FollowedCourse

prisma/schema.prisma
```diff
@@ -120,10 +123,33 @@ model User {
...
+  termId     String  @unique(map: "term_ids.term_id_unique") @map("term_id")
+  subCollege String  @map("sub_college")
   text       String
   active     Boolean @default(true)
 
   @@map("term_ids")
 }
+
+model SentNotifs {
+  userId                  String   @unique
```
mehallhm, line 135: The userId is not a string, nor would it be unique in this table (a user can get many messages). Ideally this would be a FK to the User table too so it could be joined against

prisma/schema.prisma
```diff
@@ -120,10 +123,33 @@ model User {
...
+  courseHash              String
+  subscription_created_at DateTime
+
+  timestamp DateTime
+  updateId  DateTime
+  sent      Json
+
+  Update Updates @relation(fields: [updateId], references: [updateId])
+  @@id([userId, courseHash, subscription_created_at])
```
mehallhm, line 144: I would just autoincrement an int id for this. KISS

prisma/schema.prisma
```diff
@@ -120,10 +123,33 @@ model User {
...
 }
+
+model SentNotifs {
+  userId                  String   @unique
+  courseHash              String
+  subscription_created_at DateTime
+
+  timestamp DateTime
+  updateId  DateTime
+  sent      Json
```
mehallhm, line 141: Should this just be a boolean? I don't think it could be anything other than yes it sent or no it didn't

prisma/schema.prisma
```diff
@@ -120,10 +123,33 @@ model User {
...
   @@map("term_ids")
 }
+
+model SentNotifs {
+  userId                  String   @unique
+  courseHash              String
+  subscription_created_at DateTime
+
+  timestamp DateTime
+  updateId  DateTime
```
mehallhm, line 140: Keep the timestamp, but maybe have it default to now so that way it does not need to be explicitly typed

prisma/schema.prisma
```diff
@@ -120,10 +123,33 @@ model User {
...
   active     Boolean @default(true)
 
   @@map("term_ids")
 }
+
+model SentNotifs {
+  userId                  String   @unique
+  courseHash              String
+  subscription_created_at DateTime
+
```
mehallhm, line 138: Once the userId is fixed, these fields should then be a FK to the Course / Section subscription. In Prisma, I think that would then be added in those models

mehallhm (changes requested):

mehallhm (approved):
lgtm - lets get a full team review

prisma/schema.prisma
```diff
@@ -120,10 +128,35 @@ model User {
...
+  userId                  Int
+  courseHash              String
+  subscription_created_at DateTime
+  timestamp               DateTime @default(now())
+  updateId                DateTime @updatedAt
+  sent                    Boolean
+
+  Update          Updates           @relation(fields: [updateId], references: [updateId])
+  FollowedSection FollowedSection[]
+  FollowedCourse  FollowedCourse[]
```
ItsEricSun, line 150: Maybe I'm not fully understanding how it all connects together, but why does SentNotifs have a list/one-to-many relationship with FollowedSection and FollowedCourse? Isn't a single notif for a single section/course?

prisma/schema.prisma
```diff
@@ -89,24 +89,32 @@ model Subject {
...
-  @@id([userId, courseHash])
+  courseHash String     @map("course_hash")
+  userId     Int        @map("user_id")
+  notifCount Int        @default(0) @map("notif_count")
+  created_at DateTime   @default(now())
+  updated_at DateTime   @updatedAt
+  deleted_at DateTime?
+  course     Course     @relation(fields: [courseHash], references: [id], onDelete: Cascade)
+  user       User       @relation(fields: [userId], references: [id], onDelete: Cascade)
+  SentNotifs SentNotifs @relation(fields: [userId, courseHash, created_at], references: [userId, courseHash, subscription_created_at])
```
WesleyTran0, line 100: Just making sure I understand... Every FollowedCourse and FollowedSection has a reference to the SentNotifs table? Again similar to Eric's comment on lines +149-150, wouldn't each entry in SentNotif relate to its course/section hash through the SentNotifs table?

prisma/schema.prisma
cherman23: What you said is right, I confused the syntax

mehallhm (approved):

Issue #281, opened 2025-03-20
Active terms


mehallhm:
The overall purpose of this ticket is to start having more manual control over what terms are "active" in the cca

PR #294, opened 2025-04-02
Active terms

WesleyTran0:
# Purpose

- Giving us more access on what terms are 'active'
- Notifications will only be added if they are for an 'active' class

# Tickets

#281 

# Contributors
@WesleyTran0 

# Feature List

- Added activeTerms list of strings to macro. This will be manually changed until further notice
- updater now only updates active terms. Every previous call to 'getActiveTerms' are now replaced with the recently added macros.activeTermIds
- active field is now added to termId in graphQL endpoints. This feature is reflected on the front end as well (will be mentioned in another PR to the frontend)
- prevents subscriptions to classes that aren't a part of any active terms. Filters out both section and course lists and if there are no classes to add to subscriptions after the filtering, return with a code of 204 rather than just returning 200 (may be confusing)

# Notes (Optional)

Will be connected to another PR to the front end to make the graphQL endpoint for active field accessibile on the front end

# Reviewers

Primary reviewer:

**Primary**: @mehallhm 

Secondary reviewers:

**Secondary**: @cherman23 @ItsEricSun @nickpfeiffer05 @ananyaspatil

### DIFF
--- a/graphql/typeDefs/termInfo.js
+++ b/graphql/typeDefs/termInfo.js
@@ -9,6 +9,7 @@ const typeDef = gql`
     termId: String!
     subCollege: String!
     text: String!
+    active: Boolean!
   }
 `;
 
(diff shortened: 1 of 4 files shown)


WesleyTran0:
Connected with the PR #337 in searchneu

Issue #264, opened 2025-03-16
Notification Limits on Endpoints


mehallhm:
The notification limits should be enforced in the REST endpoints. When a user goes to create a new notification, it should first check that they are under the allotted limit across **all terms**. If they are under the limit add the subscription and return a success code else return the error.

Blocked by: **TODO**

PR #295, opened 2025-04-03
Notification limit on endpoints

cherman23:
# Purpose

The limit on notifications is only enforced on the frontend. It needs to be enforced by the API as well.

# Tickets

#264 

# Contributors

@cherman23 

# Feature List

- checks for user being under limit prior to adding a subscription

Primary reviewer:

@mehallhm 
@ananyaspatil 

Secondary reviewers:

@ItsEricSun 
@nickpfeiffer05 
@WesleyTran0


cherman23:
I think I'd like to add tests actually before this is merged

